PR #3, opened 2023-11-08
Feature / notification lambda + timestream

vdektiarev:
1. Added trigger for notification lambda
2. Added Timestream DB
3. Extracted Timestream to a separate CF stack
4. Renamed infra folders and stacks.


NastassiaPaskrobka (approved):

infra_monitoring_account/tests/unit/test_infra_monitoring_stack.py
```diff
@@ -1,7 +1,7 @@
 import aws_cdk as core
 import aws_cdk.assertions as assertions
 
-from infra_monitoring.infra_monitoring_stack import InfraMonitoringStack
+from infra_monitoring_account.infra_monitoring_stack import InfraMonitoringStack
```
natmasslove, line 4: Maybe, it'd be better to change name to "infra_monitorED_account" (both, folder and stack names) - as we discussed earlier.

infra_tooling_account/infra_tooling_account/infra_tooling_alerting_stack.py
```diff
@@ -45,25 +51,40 @@ def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
             event_bus_name=f"eventbus-{project_name}-notification-{stage_name}"
         )
 
+        # Notification Lambda Eventbridge rule
+        notification_lambda_event_rule = events.Rule(
```
natmasslove, line 55: 1. I believe this rule should be in monitored account (where events are generated)
2. let's make pattern more specific
  aws:glue, ... <- services we CAN handle (+ step functions, + (?) lambda if it's possible, + (?) glue workflow if it's separate from glue itsleft)

infra_monitoring_account/tests/unit/test_infra_monitoring_stack.py
vdektiarev: Fixed

infra_tooling_account/infra_tooling_account/infra_tooling_alerting_stack.py
vdektiarev: Discussed - leaving as it should be

PR #7, opened 2023-11-15
Feature/notification lambda

vdektiarev:
1. Add notification lambda
2. Target both lambdas to use SNS topic as DLQ
3. Refactor stacks to use methods
4. Move settings bucket to the common stack (NEED FULL STACKS RECREATION)


natmasslove (approved):
Especially, I love CDK Stack being split into logically grouped method.

PR #8, opened 2023-11-15
Feature/tags configs cleanup

vdektiarev:
1. Removed internal error topic arn output from alerting stack (can't be two same outputs in the same app)
2. Passing tags to the stacks instead of tags applying commands duplication
3. Refactored monitored stack code to extract resources creation into methods
4. Adapted stacks to config file changes
5. Removed env admin email from sample configs, as we won't use them

### DIFF
--- a/infra_tooling_account/app.py
+++ b/infra_tooling_account/app.py
@@ -18,12 +18,14 @@ app = cdk.App()
 InfraToolingCommonStack(
     app,
     f"cf-{PROJECT_NAME}-InfraToolingCommonStack-{STAGE_NAME}",
+    tags={"project_name": PROJECT_NAME, "stage_name": STAGE_NAME},
     stage_name=STAGE_NAME,
     project_name=PROJECT_NAME,
 )
 InfraToolingAlertingStack(
     app,
     f"cf-{PROJECT_NAME}-InfraToolingAlertingStack-{STAGE_NAME}",
+    tags={"project_name": PROJECT_NAME, "stage_name": STAGE_NAME},
     stage_name=STAGE_NAME,
     project_name=PROJECT_NAME,
 )
--- a/infra_tooling_account/app.py
+++ b/infra_tooling_account/app.py
@@ -14,20 +14,24 @@ STAGE_NAME = os.environ["STAGE_NAME"]
 
 PROJECT_NAME = "salmon"
 
+TAGS = {"project_name": PROJECT_NAME, "stage_name": STAGE_NAME}
+
 app = cdk.App()
-InfraToolingCommonStack(
+common_stack = InfraToolingCommonStack(
     app,
     f"cf-{PROJECT_NAME}-InfraToolingCommonStack-{STAGE_NAME}",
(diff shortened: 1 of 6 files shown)


natmasslove (approved):

infra_tooling_account/app.py
```diff
@@ -18,12 +18,14 @@
...
     app,
     f"cf-{PROJECT_NAME}-InfraToolingCommonStack-{STAGE_NAME}",
+    tags={"project_name": PROJECT_NAME, "stage_name": STAGE_NAME},
     stage_name=STAGE_NAME,
     project_name=PROJECT_NAME,
 )
 InfraToolingAlertingStack(
     app,
     f"cf-{PROJECT_NAME}-InfraToolingAlertingStack-{STAGE_NAME}",
+    tags={"project_name": PROJECT_NAME, "stage_name": STAGE_NAME},
```
natmasslove, line 28: Alerting stack now relies on Common stack's output, so we need to make sure Alerting is created ONLY after Common is completed (I bumped into problem like this).

commonStack = InfraToolingCommonStack(
    app,
    f"cf-{PROJECT_NAME}-InfraToolingCommonStack-{STAGE_NAME}",
    tags={"project_name": PROJECT_NAME, "stage_name": STAGE_NAME},
    stage_name=STAGE_NAME,
    project_name=PROJECT_NAME,
)
infraStack = InfraToolingAlertingStack(
    app,
    f"cf-{PROJECT_NAME}-InfraToolingAlertingStack-{STAGE_NAME}",
    tags={"project_name": PROJECT_NAME, "stage_name": STAGE_NAME},
    stage_name=STAGE_NAME,
    project_name=PROJECT_NAME,
)

infraStack.add_dependency(commonStack)

infra_tooling_account/infra_tooling_account/infra_tooling_alerting_stack.py
natmasslove, line 1: Another small issue - eventbusrule-salmon-alerting-lambda-{stage} is not getting tagged.
I guess, we need to employ custom tagging for this specific resource

natmasslove: Looks like CloudFormation's shortcoming. No actions needed.
https://github.com/aws/aws-cdk/issues/4907

infra_tooling_account/app.py
```diff
@@ -18,12 +18,14 @@
 InfraToolingCommonStack(
     app,
     f"cf-{PROJECT_NAME}-InfraToolingCommonStack-{STAGE_NAME}",
+    tags={"project_name": PROJECT_NAME, "stage_name": STAGE_NAME},
```
natmasslove, line 21: Maybe, it's better to define tags var once
tags={"project_name": PROJECT_NAME, "stage_name": STAGE_NAME}
and use it in all stack's params?

infra_tooling_account/app.py
vdektiarev: Done

infra_tooling_account/app.py
vdektiarev: Done

PR #10, opened 2023-11-20
Feature/notification service

nalkhimo:
- Added AwsSesSender, SmtpSender


src/constants.py
```diff
@@ -2,3 +2,9 @@ class SettingFileNames:
     GENERAL_FILE_NAME = "general.json"
     MONITORING_GROUPS_FILE_NAME = "monitoring_groups.json"
     RECIPIENTS_FILE_NAME = "recipients.json"
+
+
+class NotificationServiceConfig:
```
natmasslove, line 7: I'm not sure if we need this specific values here.
If it's specific to lambda_test_notification_service - let's place those inside lambda code.
If not - let's discuss:
alert_header (or generally speaking "message subject", right?) - should be passed from a lambda which calls notification lambda (either alerting lambda or digest lambda).

src/constants.py
```diff
@@ -2,3 +2,9 @@ class SettingFileNames:
...
+    SES_SENDER = "my_email@soname.de"
+    ALERT_HEADER = "SDM [DEV] SAP Lean Glue Failure"
+    SMTP_SENDER = "my_email@soname.de"
```
natmasslove, line 10: for ses &smtp sender - it's a good point, we need to put it somewhere, but more proper way to define it in general.json config.
Like:

    "delivery_methods": [
        {
            "name": "aws_ses",
            "delivery_method_type": "AWS_SES",
            "sender_email" : "salmon-no-reply@soname.de"
        },

src/notification_service/sender/ses.py
```diff
@@ -0,0 +1,151 @@
...
+    def verified_recipients(self) -> List[str]:
+        """Verified by AWS SES recipiens."""
+        return self._verified_recipients
+
+    @verified_recipients.setter
+    def verified_recipients(self, recepients: List[str]) -> None:
+        """Verified by AWS SES recipiens."""
+        self._verified_recipients = recepients
+
+    def _get_identities(self, next_token: str = None) -> List[str]:
```
natmasslove, line 45: I would prefer to move this functionality into another lib - aws-service related (we have s3-Manager already, it looks like we can also have ses_manager module with 1 function is_email_verified)

src/notification_service/sender/base.py
```diff
@@ -0,0 +1,21 @@
+from abc import ABC, abstractmethod
```
NastassiaPaskrobka, line 1: Wouldn't it be better to name class and file with the same name? What do you think?

NastassiaPaskrobka (commented):
Really like an approach with sender provider and registration of the recipients! Nice architecture for this use-case 👍

src/constants.py
nalkhimo: Agree. At the beginning of development, I thought it would be located in the constants section, but then we discussed that it is not necessary and all these variables will be passed to this lambda.

src/notification_service/sender/base.py
nalkhimo: I called it base just to emphasize that this class is a base class. But I can rename it as well.

vdektiarev (approved):

src/notification_service/sender/base.py
NastassiaPaskrobka: I've seen from some best practises that if a module defines a single class, the module name and class name should match. It increases readability when you want to find a code of the class.
Like your example:
class name = SenderProvider
module_name = sender_provider.py

If you do not mind, let's use consistent naming throughout the code where we have one module = one class :)

nalkhimo: Thank you for the research, I didn't know that. I will change the name then 👍

natmasslove (approved):

PR #11, opened 2023-11-20
Feature/settings validation

NastassiaPaskrobka:
Settings validation added (both business and schema validation)


config/test_settings/general.json
```diff
@@ -0,0 +1,30 @@
+{
```
natmasslove, line 1: I'm not sure if we want to put actual configs into main repo

tests/src/lib/aws/test_s3_manager.py
```diff
@@ -1,7 +1,7 @@
 import pytest
 from botocore.exceptions import ClientError
 from unittest.mock import MagicMock
-from src.lib.s3_manager import S3Manager, S3ManagerReadException
+from src.lib.aws import S3Manager, S3ManagerReadException
```
natmasslove, line 4: I like this separation! :+1:

natmasslove (approved):
Very good foundation for settings usage and more future improvements!

vdektiarev:
No other comments except for removing test_settings folder along with the config files, as they are in testing_stand now. Everything else looks good!

src/requirements.txt
```diff
@@ -0,0 +1,2 @@
+boto3  # AWS SDK for Python
+jsonschema # Validate JSON settings schema
```
vdektiarev, line 2: I'm afraid there's no such package available in the default Lambda runtime: https://gist.github.com/gene1wood/4a052f39490fae00e0c3

We can either use something mentioned in the gist above, or use Docker to prepare a layer with all packages from requirements.txt

natmasslove: I agree. 
We have just one use-case where we use configs validation - inside CDK code, before deployment we want to make sure
a) configs are present (where we expect to find them :) )
b) configs are valid.
So it tells us, we can deploy this version of configs to S3 and safely use it.

Manually changing configs directly on S3 is not good scenario for too many reasons :) 
So, deploy configs only via CDK process

vdektiarev: @NastassiaPaskrobka , may I ask you then to extract settings validation functionality into a separate module so that we would use it directly in CDK only? In that way Lambdas will use only settings readers, and not settings validation, and won't have jsonschema import issues.

NastassiaPaskrobka: Sure, makes sense, I'll move it to infra_tooling_stack then

NastassiaPaskrobka: Just realized that Validator uses settings reader also.
So once again open question where to put settings reader to be commonly accessed by lambdas and CDK :)

src/lambda/test-settings-lambda/index.py
```diff
@@ -1,10 +1,70 @@
...
+        jsonschema.exceptions.ValidationErro: If any schema validation checks fail.
+        SettingsValidatorException: If any business validation checks fail.
+
+    """
+    schema_base_path = "schemas/"
+    config_base_path = "../config/test_settings/"
+
+    general = GeneralSettingsReader(
+        SettingFileNames.GENERAL_FILE_NAME,
+        fm.read_file(config_base_path + "general.json"),
```
nalkhimo, line 38: Better to use os.path.join() for creating a path.

nalkhimo (changes requested):
Great job as usual :)

nalkhimo (approved):

src/lib/core/file_manager.py
```diff
@@ -0,0 +1,58 @@
+# Exceptions
+class FileManagerReadException(Exception):
+    """Exception raised for errors during file reading."""
+
+    def __init__(self, message):
```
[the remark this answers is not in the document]
NastassiaPaskrobka: Yeap, actually it does not change the Exception functionality, just thought it is a better practice to identify at least something in class, but we can avoid this. Let's agree just to put everythere
class ExceptionClass(Exception):
    pass 
    
I'll change it in all the remaining places also :)

src/lib/settings/settings_reader/recipients_settings_reader.py
```diff
@@ -36,11 +36,11 @@ def get_monitoring_group_names(self) -> list[str]:
...
-        monitoring_group_names = []
-
-        for recipient in self._recipients:
-            for subscription in recipient.get("subscriptions"):
-                monitoring_group_names.append(subscription.get("monitoring_group"))
+        monitoring_group_names = [
+            subscription.get("monitoring_group")
+            for recipient in self._recipients
+            for subscription in recipient.get("subscriptions")
+        ]
```
natmasslove, line 43: let's keep method as it is for now (this PR).
Later we need to review some methods.
For example if this is for getting recipient's subscriptions to send alert or digest - than it requires additional parameter ( subscription_type : alerts / digest ) and filter only when subscription_type is true in config

natmasslove (approved):

src/lib/settings/settings_reader/recipients_settings_reader.py
NastassiaPaskrobka: yes, currently this method is specially for the validation of all the recepients config.
Probably we would need to make it more generic to provide also the functionality you describing
P.S. For the specific type  currently there is alother method used - get_recipients_by_monitoring_groups_and_type

PR #13, opened 2023-11-21
Feature/alerting lambda to timestream

natmasslove:


infra_tooling_account/infra_tooling_account/infra_tooling_common_stack.py
```diff
@@ -96,6 +116,21 @@ def create_timestream_db(self):
 
         return timestream_storage
 
+    def create_timestream_tables(self, timestream_storage):
+        retention_properties_property = timestream.CfnTable.RetentionPropertiesProperty(
+            magnetic_store_retention_period_in_days="365",
```
nalkhimo, line 121: Wouldn't be better to move these contants "365" and "240" into settings or contants.py file?

src/lib/aws/timestream_manager.py
```diff
@@ -0,0 +1,85 @@
...
+        # Convert the ISO date string to a datetime object
+        dt = datetime.fromisoformat(iso_date.rstrip("Z"))
+        # Convert the datetime object to epoch time in seconds
+        epoch_time = dt.timestamp()
+
+    # Convert epoch time to milliseconds and return as string
+    return str(int(epoch_time * 1000))
+
+
+class TimestreamTableWriter:
```
nalkhimo, line 33: What do you think about renaming this class to TimestreamManager? And in future we will be able to add new methods connected to Timestream if needed.

src/lib/aws/timestream_manager.py
```diff
@@ -0,0 +1,85 @@
...
+        Args:
+            records: A list of records to be written to the Timestream table.
+
+        Returns:
+            The response from the Timestream write_records API call.
+
+        Todo:
+            Introduce records buffering to support batches of less than 100 records.
+        """                
+        result = self.timestream_write_client.write_records(
```
nalkhimo, line 80: Better to add try-except block with a custom exception. That will help in future to understand where error occurred.

nalkhimo (approved):

NastassiaPaskrobka (approved):

infra_tooling_account/infra_tooling_account/infra_tooling_common_stack.py
natmasslove: Agree. Added trello task for future not to bloat current PR

src/lib/aws/timestream_manager.py
natmasslove: Actually, that was the idea to name like that.
In future we will have, I guess, TimestreamTableReader, which is significantly different.
It even uses another boto3 client (there are separate clients for timestream write and read).

src/lib/aws/timestream_manager.py
natmasslove: fixed

vdektiarev (approved):

infra_tooling_account/infra_tooling_account/infra_tooling_alerting_stack.py
```diff
@@ -150,7 +166,7 @@ def create_alerting_lambda(
 
         alerting_lambda_role.add_to_policy(
             iam.PolicyStatement(
-                actions=["queue:SendMessage"],
```
vdektiarev, line 153: My bad, looked in the wrong place for permission names references. Good catch!

PR #14, opened 2023-11-22
Feature/lambda location

vdektiarev:
1. Move lambdas to src directory
2. Use settings reader

Can wait for you guys to merge your parts :)


infra_tooling_account/infra_tooling_account/infra_tooling_alerting_stack.py
```diff
@@ -77,22 +81,17 @@ def create_event_bus(self):
         )
 
         # EventBridge bus resource policy
-        # TODO: reuse existing settings reader
+        # TODO: use settings validation
```
natmasslove, line 84: another approach would be to read and validate all settings just once - in cdk app.py before stacks.
After that we consider all settings as valid.

Not for this PR specifically - but for future, when we finalize settings classes structure

src/lib/constants.py
```diff
@@ -2,3 +2,7 @@ class SettingFileNames:
     GENERAL_FILE_NAME = "general.json"
     MONITORING_GROUPS_FILE_NAME = "monitoring_groups.json"
     RECIPIENTS_FILE_NAME = "recipients.json"
+
+
+class Exclusions:
```
natmasslove, line 7: do you think it makes sense to name it CDKDeployExclusions?

constants.py is just a general-purpose file, so, maybe, name needs to be a bit more specific?

natmasslove (approved):

infra_tooling_account/infra_tooling_account/infra_tooling_monitoring_stack.py
```diff
@@ -120,11 +124,13 @@ def get_metrics_collection_interval_min(self):
...
         with open(general_settings_file_path) as f:
-            general_config = json.load(f)
-            tooling_section = general_config["tooling_environment"]
+            general_settings = settings_reader.GeneralSettingsReader(
+                general_settings_file_path, f.read()
+            )
+            tooling_section = general_settings.get_tooling_environment()
             key = "metrics_collection_interval_min"
             if key in tooling_section:
                 return tooling_section[key]
```
NastassiaPaskrobka, line 136: Here the same regarding access to settings values :)

infra_monitored_account/infra_monitored_account/infra_monitored_stack.py
```diff
@@ -46,8 +43,9 @@ def create_cross_account_event_bus_role(self):
             assumed_by=iam.ServicePrincipal("events.amazonaws.com"),
         )
 
-        tooling_account_id = general_config["tooling_environment"]["account_id"]
-        tooling_account_region = general_config["tooling_environment"]["region"]
+        tooling_environment = general_settings.get_tooling_environment()
+        tooling_account_id = tooling_environment["account_id"]
+        tooling_account_region = tooling_environment["region"]
```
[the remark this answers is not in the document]
vdektiarev: Yeah, there's a Trello card for that: https://trello.com/c/7XVJHNkp

src/lib/constants.py
vdektiarev: Done

infra_tooling_account/infra_tooling_account/infra_tooling_alerting_stack.py
```diff
@@ -77,22 +81,17 @@ def create_event_bus(self):
...
-            try:
-                general_config = json.load(f)
-            except json.decoder.JSONDecodeError as e:
-                raise json.decoder.JSONDecodeError(
-                    f"Error parsing JSON file {general_settings_file_path}",
-                    e.doc,
-                    e.pos,
-                )
+            general_settings = settings_reader.GeneralSettingsReader(
+                general_settings_file_path, f.read()
```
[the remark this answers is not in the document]
vdektiarev: Adapted

infra_monitored_account/app.py
```diff
@@ -1,10 +1,57 @@
...
+from lib.core.constants import SettingFileNames
+from lib.settings.settings_reader import (
+    GeneralSettingsReader,
+    MonitoringSettingsReader,
+    RecipientsSettingsReader,
+)
+from lib.settings import settings_validator
+
+
+def read_settings():
```
vdektiarev, line 20: Can leave it here or create another utility module in src/settings to have something that can accept files path, read and validate all files from there. Do you think it's a good idea?

infra_tooling_account/app.py
```diff
@@ -17,6 +66,12 @@
 
 TAGS = {"project_name": PROJECT_NAME, "stage_name": STAGE_NAME}
 
+(
+    general_settings_reader,
+    monitoring_groups_settings_reader,
+    recipients_settings_reader,
+) = read_settings()
+
```
natmasslove, line 74: let's keep it like this for the current PR.
In future, I'd say:

settings_reader = LocalFSSettingsReader(config_folder)
settings = settings_reader.read_and_prepare()
settings.validate()

and, then, settings (which is an object containing all 3 settings structs) are passed to each stack construct as a param

@NastassiaPaskrobka what do you think?

vdektiarev: I had an idea of creating settings_provider module and putting it along with settings_validator under src/lib/settings/cdk folder - to make sure this would be used by CDK only.

And there i declared the same read function (naming can be discussed) which accepts config folder path, reads and validates all the files - and provides settings readers as the result. Tested this approach - works fine, and no such read_settings functions code duplication.

vdektiarev: `(
    general_settings_reader,
    monitoring_groups_settings_reader,
    recipients_settings_reader,
) = settings_provider.read_settings("../config/settings")`

That's all it takes to call it in app.py

aryna-yukhnevich (approved):

PR #15, opened 2023-11-27
Added formatter

nalkhimo:
- new: Added formatter


src/lambda_test_notification_service_formatter.py
```diff
@@ -0,0 +1,59 @@
+import json
+from lib.notification_service.formatter_provider import formatters
+
+
+def lambda_handler(event, context):
+    event_info = json.loads(event)
+    delivery_options_info = event_info.get("delivery_options")
```
natmasslove, line 7: maybe it also makes sense to check presence of data and delivery_options in event.

src/lib/notification_service/formatter_provider.py
```diff
@@ -0,0 +1,24 @@
...
+
+        if not formatter:
+            raise ValueError(f"Delivery method {delivery_method} is not supported.")
+
+        return formatter()
+
+
+formatters = FormatterProvider()
+formatters.register_sender("AWS_SES", HtmlFormatter)
+formatters.register_sender("SMTP", HtmlFormatter)
```
npaskrobka, line 24: Just though if it possible to have just one provider for both Sender and Formatter? Like 
```
provider = NotificationProvider()
provider.register_sender("AWS_SES", create_aws_ses_sender, HtmlFormatter)
```
It is just an idea since we have both formatter and sender per each delivery method, but not sure if it possible in such case :)
Everything else looks great, as usual!

npaskrobka (approved):

aryna-yukhnevich (approved):

src/lib/notification_service/formatter/htmp_formatter.py
vdektiarev, line 1: Looks like a typo in the filename

src/lib/notification_service/formatter_provider.py
```diff
@@ -0,0 +1,24 @@
+from .formatter import HtmlFormatter
+
+
+class FormatterProvider:
+    """Formatter factory."""
+
+    def __init__(self):
+        self._formatters = {}
+
+    def register_sender(self, delivery_method, sender):
```
vdektiarev, line 10: Is _register_sender_ going to be called anywhere else besides this class? If not, why is it not preferable to initialize the dictionary of formatters in the constructor? The reason I'm asking is because I'm doing kind of the same with glue/stepfunctions event parsers :)

src/lib/notification_service/formatter_provider.py
nalkhimo: Thanks for the comment, but I believe it is better not to mix these 2 different entities, because get method and register_sender should retrieve and update different dictionaries.

PR #17, opened 2023-11-28
Refactored settings component

npaskrobka:
1. Concepts "raw settings" and "processed settings"
      * raw - as they are stored in git, on S3 or file system
      * processed - based on the read raw, we open wildcards and add defaults
2. Use-cases:
2.1 (in CDK) validation of configs (json structure + business rules)
2.2 (in CDK) methods
     * get_monitored_account_ids
     * get_metrics_collection_interval_min (from tooling_env)
2.3 (in lambda) methods
     * get_monitored_env_name(account_id, region)
     * get_relevant_monitoring_groupS ( monitored item, e.g. glue job )
     * get_recipients (monitoring_group, alert/digest)
     and similar
3. Raw -> Processed for monitoring groups.json can be an expensive operation - done in a lazy way with cached_property, only be puzzled by this when we are actually asked about monitoring groups
​
Raw -> Processed:
image
image


src/lib/aws/s3_manager.py
```diff
@@ -32,21 +32,23 @@ def __init__(self, s3_client: boto3.client = None):
...
-            file_name (str): The name of the file to read.
+            s3_path (str): Full S3 path (e.g. s3://your_bucket_name/path/to/your/object/file.txt).
 
         Returns:
             str: The content of the file as a string or None if an error occurs.
 
         """
         try:
-            response = self.s3_client.get_object(Bucket=bucket_name, Key=file_name)
+            s3_path_parts = s3_path.replace("s3://", "").split("/", 1)
```
natmasslove, line 46: can we use urlparse lib?
https://stackoverflow.com/questions/42641315/s3-urls-get-bucket-name-and-path

the lib is available in lambda

src/lib/settings/settings.py
```diff
@@ -0,0 +1,280 @@
...
+        ]
+
+    def get_metrics_collection_interval_min(self) -> int:
+        """Get metrics_collection_interval_min"""
+        return self.processed_settings[SettingFileNames.GENERAL]["tooling_environment"][
+            "metrics_collection_interval_min"
+        ]
+
+    # Lambda methods
+    def get_monitored_environment_name(self, account_id: str, region: str) -> List[str]:
```
natmasslove, line 133: I guess this method shouldn't return list, but rather a single value.

account_id + region combination uniquely identifies monitored env

src/lib/settings/settings.py
```diff
@@ -0,0 +1,280 @@
...
+    def get_monitored_account_id_and_region_raw_gs(self) -> List[str]:
+        """Retrieves the monitored environments 'account_id|region' from the General settings (raw)."""
+        return [
+            f"""{m_env["account_id"]}|{m_env["region"]}"""
+            for m_env in self._raw_settings[SettingFileNames.GENERAL].get(
+                "monitored_environments", []
+            )
+        ]
+
+    def get_delivery_method_names_raw_gs(self) -> List[str]:
```
natmasslove, line 206: Maybe, it's better not to have any get method which rely on _raw_settings.

So far, we don't have "processed" for monitoring groups, but we can pretend "raw" state = "processed" stage for now. -> we can make these "get_" methods final, accessing processed settings.
Later - implement this raw -> processed transformations is the only change needed

src/lib/settings/settings.py
```diff
@@ -0,0 +1,280 @@
...
+        recipients_settings = fm.read_file(
+            os.path.join(base_path, SettingFileNames.RECIPIENTS)
+        )
+        return cls(general_settings, monitoring_settings, recipients_settings)
+
+    @classmethod
+    def from_s3_path(cls, base_path: str) -> object:
+        s3 = S3Manager()
+        general_settings = s3.read_file(
+            f"{base_path.rstrip('/')}/{SettingFileNames.GENERAL}"
```
natmasslove, line 272: can we use os.path.join() here?

natmasslove, line 272: can we use os.path.join here as well?

>>> os.path.join('s3://mybucket/foo/bar', 'add/me')
's3://mybucket/foo/bar/add/me'
>>> os.path.join('s3://mybucket/foo/bar/', 'add/me')
's3://mybucket/foo/bar/add/me'
https://stackoverflow.com/questions/49078156/use-pathlib-for-s3-paths

npaskrobka: done

src/lib/settings/settings.py
npaskrobka: done

src/lib/aws/s3_manager.py
npaskrobka: done

src/lib/settings/cdk/settings_validator.py
```diff
@@ -4,7 +4,7 @@
```
natmasslove, line 4: when I try to validate settings in app.py (code below) :

FileNotFoundError: File not found: src/lib/settings/cdk/schemas/general.json

# settings_provider = LocalFileSystemSettingsProvider("../config/settings")
# settings_provider.validate_settings()

current_settings = Settings.from_file_path("../config/settings")
settings_validator.validate(current_settings)

natmasslove (approved):
Approved, keeping in mind an error which we get while validating setting in app.py.
Can be fixed as a separate PR

src/lib/settings/settings.py
```diff
@@ -0,0 +1,214 @@
...
+        """Get recipients by monitoring groups."""
+        matched_recipients = []
+
+        for recipient in self.recipients.get("recipients", []):
+            subscriptions = recipient.get("subscriptions", [])
+            for subscription in subscriptions:
+                for monitoring_group in monitoring_groups:
+                    if subscription.get("monitoring_group") == monitoring_group:
+                        if (
+                            notification_type == "alert" and subscription.get("alerts")
```
vdektiarev, line 173: Maybe it would be better to create a class with constants (enum) for that? Would be more user-friendly

src/lib/settings/cdk/settings_validator.py
npaskrobka: fixed: added full path as discussed :)

src/lib/settings/settings.py
npaskrobka: Good notice! Done

vdektiarev (approved):

PR #21, opened 2023-12-04
Create MonitoredAcc Role for ExtractMetrics + Refactor CDK Stacks to use new Settings and validator

natmasslove:
1. Added an IAM Role which permit tooling env (Extract Metrics Lambda) to read monitored's env data (currently: Glue and Lambda)
2. Added sample code to get cross-account Glue Job list from the secondary testing stand
  You'll need to update monitored stacks in your env (second account)
3. Renamed IAM Roles in monitored acc (so we can clearly distinguish those belonging to monitored stack)
4. Refactored CDK Code for both, tooling and monitored, to use the latest Settings class and settings validator


aryna-yukhnevich (approved):

infra_tooling_account/app.py
```diff
@@ -26,8 +26,9 @@
 
 TAGS = {"project_name": PROJECT_NAME, "stage_name": STAGE_NAME}
 
-settings_provider = LocalFileSystemSettingsProvider("../config/settings")
-settings_provider.validate_settings()
+settings = Settings.from_file_path("../config/settings")
+settings_validator.validate(settings)
```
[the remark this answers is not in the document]
vdektiarev: I guess it would depend on a user, what would be installed in the landing zone first: tooling or monitoring parts. And I believe it's better to validate settings as soon as possible. I would prefer leaving validations in both stacks just in case.

infra_monitored_account/infra_monitored_account/infra_monitored_stack.py
```diff
@@ -82,3 +85,52 @@ def create_event_rules(self, cross_account_bus_role, cross_account_event_bus_arn
...
+        principal_arn = AWSNaming.Arn_IAMRole(self, self.tooling_account_id, tooling_extract_lambda_role_name)
+        
+        # todo rename
+        cross_account_iam_role_extract_metrics = iam.Role(
+            self, "MonitoredAccExtractMetricsRole",
+            role_name=AWSNaming.IAMRole(self, CDKResourceNames.IAMROLE_MONITORED_ACC_EXTRACT_METRICS),            
+            assumed_by=iam.ArnPrincipal(principal_arn)
+        )
+
+        # Glue Policy
```
vdektiarev, line 105: Not going to add Step Functions policy for now?

infra_tooling_account/infra_tooling_account/infra_tooling_common_stack.py
```diff
@@ -19,12 +19,14 @@
...
 from lib.core.constants import CDKDeployExclusions, TimestreamRetention
 from lib.aws.aws_naming import AWSNaming
+from lib.settings.settings import Settings
 
 
 class InfraToolingCommonStack(Stack):
     def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
         self.stage_name = kwargs.pop("stage_name", None)
         self.project_name = kwargs.pop("project_name", None)
+        self.settings: Settings = kwargs.pop("settings", None)
```
vdektiarev, line 29: Do we need settings component in common stack? Maybe it's better to pass it only when needed?

src/lib/core/constants.py
```diff
@@ -23,3 +23,14 @@ class NotificationType:
...
 class CDKDeployExclusions:
     LAMBDA_ASSET_EXCLUSIONS = [".venv/", "__pycache__/"]
+
+class CDKResourceNames:
+    """ Contains 'meaningful' part of AWS resources names. 
+    Specifically, ones referred in both, tooling and monitored accounts
+    """
+    IAMROLE_EXTRACT_METRICS_LAMBDA = "extract-metrics-lambda"
+    IAMROLE_MONITORED_ACC_PUT_EVENTS = "monitored-acc-put-events"
+    IAMROLE_MONITORED_ACC_EXTRACT_METRICS = "monitored-acc-extract-metrics"
```
vdektiarev, line 33: Great idea about CDK resource names, it would also be useful to add Event Bus name there too, as we are working with it both in monitored and tooling stacks. We can do this later.

infra_tooling_account/infra_tooling_account/infra_tooling_common_stack.py
vdektiarev: Please disregard - viewed the PR till the end 🙃

src/lib/settings/settings.py
```diff
@@ -116,19 +116,24 @@ def get_raw_settings(self, file_name: str) -> dict:
     # CDK methods
     def get_monitored_account_ids(self) -> List[str]:
```
vdektiarev, line 117: May only want to change the return type

infra_tooling_account/infra_tooling_account/infra_tooling_alerting_stack.py
```diff
@@ -92,14 +93,10 @@ def create_event_bus(self):
 
         # EventBridge bus resource policy
         monitored_account_ids = sorted(
-            set(
```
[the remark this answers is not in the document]
vdektiarev: Please disregard - read the PR till the end 🙃

vdektiarev (approved):

infra_tooling_account/app.py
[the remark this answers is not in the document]
natmasslove: @npaskrobka , let's discuss your idea separately from this PR (maybe, during the next sync-up). It looks like I didn't fully understand it.

Another idea worth a separate discussion - forcing user to create tooling infra first before creating corresponding monitored infra. I like your idea, but we can achieve it differently maybe :) Let's brainstorm

infra_monitored_account/infra_monitored_account/infra_monitored_stack.py
natmasslove: Yes, I wanted to speed up this PR finalization (as it touches all the CDK stacks and creates potential git conflict)
and defer it till the moment we define metrics required for Step Functions.

infra_tooling_account/infra_tooling_account/infra_tooling_common_stack.py
natmasslove: Actually, I did that on purpose - in case we'd need settings in common stack sooner or later and, more importantly, for unification (as all stacks' initialization works in the same way).

Just as a thought - we might even consider moving stage, project, settings init into SalmonStack parent class and, then, inherit from SalmonStack, not just CDK Stack

src/lib/settings/settings.py
natmasslove: What return type would you consider as preferable?

src/lib/core/constants.py
natmasslove: Nice catch. Fully agree. Fixed.

src/lib/settings/settings.py
vdektiarev: set[str]?

PR #22, opened 2023-12-04
Settings: Added replacements.json

npaskrobka:
1. Added replacements.json (as optional file).
2. Added replacements to the processed_settings.


natmasslove (approved):
It's very good you made this file optional!

src/lib/settings/settings.py
```diff
@@ -187,30 +219,59 @@ def get_recipients(
...
-            read_file_func(os.path.join(base_path, file_name))
-            for file_name in file_names
-        ]
+        settings = []
+        for file_name in file_names:
+            try:
+                file_content = read_file_func(os.path.join(base_path, file_name))
+                settings.append(file_content)
+            except Exception as e:
+                if file_name == SettingFileNames.REPLACEMENTS:
```
vdektiarev, line 228: What if any other important exception occurs during replacements processing? Wouldn't that make troubleshooting process more confusing as we are silenting all exceptions during read_file_func execution?

npaskrobka: Good remark! Changed to skip only FileNotFound error for replacements.json

vdektiarev (approved):

PR #23, opened 2023-12-04
Feature - Alerting Lambda

vdektiarev:
Alerting Lambda implementation + usage of Settings component in the tooling stack

### DIFF
--- a/infra_tooling_account/infra_tooling_account/infra_tooling_alerting_stack.py
+++ b/infra_tooling_account/infra_tooling_account/infra_tooling_alerting_stack.py
@@ -200,7 +200,7 @@ class InfraToolingAlertingStack(Stack):
             runtime=lambda_.Runtime.PYTHON_3_11,
             environment={
                 "SETTINGS_S3_BUCKET_NAME": settings_bucket.bucket_name,
-                "NOTIFICATION_QUEUE_NAME": notification_queue.queue_name,
+                "NOTIFICATION_QUEUE_URL": notification_queue.queue_url,
                 "ALERT_EVENTS_DB_NAME": timestream_database_name,
                 "ALERT_EVENTS_TABLE_NAME": timestream_alert_events_table_name,
             },
--- a/infra_tooling_account/infra_tooling_account/infra_tooling_alerting_stack.py
+++ b/infra_tooling_account/infra_tooling_account/infra_tooling_alerting_stack.py
@@ -148,7 +148,7 @@ class InfraToolingAlertingStack(Stack):
             iam.PolicyStatement(
                 actions=["s3:GetObject"],
                 effect=iam.Effect.ALLOW,
-                resources=[settings_bucket.bucket_arn],
+                resources=[f"{settings_bucket.bucket_arn}/*"],
             )
         )
 
--- a/infra_tooling_account/infra_tooling_account/infra_tooling_alerting_stack.py
+++ b/infra_tooling_account/infra_tooling_account/infra_tooling_alerting_stack.py
@@ -123,13 +123,13 @@ class InfraToolingAlertingStack(Stack):
 
     def create_alerting_lambda(
         self,
-        settings_bucket,
-        notification_queue,
-        internal_error_topic,
-        timestream_database_arn,
-        timestream_database_name,
-        timestream_alert_events_table_name,
-        alerting_lambda_event_rule,
+        settings_bucket: s3.Bucket,
+        notification_queue: sqs.Queue,
+        internal_error_topic: sns.Topic,
+        timestream_database_arn: str,
+        timestream_database_name: str,
+        timestream_alert_events_table_name: str,
+        alerting_lambda_event_rule: events.Rule,
     ):
         alerting_lambda_role = iam.Role(
             self,
@@ -199,7 +199,7 @@ class InfraToolingAlertingStack(Stack):
             timeout=Duration.seconds(120),
             runtime=lambda_.Runtime.PYTHON_3_11,
             environment={
-                "SETTINGS_S3_BUCKET_NAME": settings_bucket.bucket_name,
+                "SETTINGS_S3_PATH": f"s3://{settings_bucket.bucket_name}/settings/",
                 "NOTIFICATION_QUEUE_URL": notification_queue.queue_url,
                 "ALERT_EVENTS_DB_NAME": timestream_database_name,
                 "ALERT_EVENTS_TABLE_NAME": timestream_alert_events_table_name,
--- a/infra_tooling_account/infra_tooling_account/infra_tooling_alerting_stack.py
+++ b/infra_tooling_account/infra_tooling_account/infra_tooling_alerting_stack.py
@@ -20,7 +20,31 @@ from lib.settings.settings import Settings
 
 
 class InfraToolingAlertingStack(Stack):
+    """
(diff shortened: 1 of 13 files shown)


infra_tooling_account/infra_tooling_account/infra_tooling_alerting_stack.py
```diff
@@ -22,7 +22,7 @@ class InfraToolingAlertingStack(Stack):
     def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
         self.stage_name = kwargs.pop("stage_name", None)
         self.project_name = kwargs.pop("project_name", None)
-        self.general_settings_reader = kwargs.pop("general_settings_reader", None)
+        self.settings = kwargs.pop("settings", None)
```
natmasslove, line 25: We have a lot of changes duplicated in our PRs, inevitably.
This is the one of those.

Here, we might want to change  this piece of code to:
self.settings: Settings = kwargs.pop("settings", None)

It helps intellisense a lot, when you start using IDE to call settings methods

src/lambda_alerting.py
```diff
@@ -39,48 +43,48 @@ def write_event_to_timestream(records):
...
-        account (str): AWS account ID.
-        region (str): AWS region name.
+        s3_bucket_name (str): Name of the Settings S3 bucket
 
     Returns:
-        str: A string representing the name of the monitored environment.
+        Settings: Setting container
     """
-    return "monitored_env_from_stub"
+    settings_path = f"s3://{s3_bucket_name}/settings/"
```
natmasslove, line 55: I believe, it's better to pass the full S3 path as env variable into lambda from CDK.
instead of just passing bucketname

src/lambda_alerting.py
```diff
@@ -118,48 +122,104 @@ def prepare_timestream_record(event_props, event):
...
+    settings_bucket_name = os.environ["SETTINGS_S3_BUCKET_NAME"]
+    settings = read_settings(settings_bucket_name)
 
-    # 1. parses event to identify "source" - e.g. "glue"
+    monitored_env_name = get_monitored_env_name(event, settings)
 
-    # 2. Create instance of <source>AlertParser class and calls method "parse"
-    # parse method returns DSL-markedup message
+    messages = map_to_notification_messages(event, settings)
+    print(messages)
```
natmasslove, line 150: should we switch this to logger.info?

src/lib/event_mapper/impl/glue_event_mapper.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def get_message_body(self, event):
+        message_body = []
+        table = {}
+        rows = []
+        table["table"] = {}
+        table["table"]["rows"] = rows
+        message_body.append(table)
+
+        style = "error" if event["detail"]["severity"] == "ERROR" else None
```
natmasslove, line 20: just curious - if we have State = "STOPPED" (meaning it was stopped manually, which is kindof an error) - will it still be severity == ERROR?

Also, I believe, going forward, we'd need to discuss how to distinguish notifications about successful executions and failures

src/lib/event_mapper/impl/general_aws_event_mapper.py
```diff
@@ -0,0 +1,146 @@
...
+            event (dict): Event object
+
+        Returns:
+            str: Message subject
+        """
+        monitored_env_name = self.settings.get_monitored_environment_name(
+            event["account"], event["region"]
+        )
+        event_message = event["detail-type"]
+        return f"{monitored_env_name}: {event_message}"
```
natmasslove, line 112: It's not critical to changed it in this PR (maybe, as a follow-up task), but I believe we need subject line to be a bit informative.
Instead of "Env1: Glue Job State Change" it's better to have
"Env1: Glue Job my-lovely-job has FAILED" (SUCCEEDED).

For users it'd be much easier to set up e-mail filters etc.

src/lib/event_mapper/impl/glue_event_mapper.py
vdektiarev: Good question, I would believe that it wouldn't have "error" severity, as it's not listed as "abnormal" run status for jobs, so only "FAILED", "ERROR" and "TIMEOUT" would be treated as something not good happening: https://docs.aws.amazon.com/glue/latest/dg/job-run-statuses.html

However we can discuss how to better determine the severity of events 👍 

Agree, this one's an initial implementation. Created a card in Trello.

infra_tooling_account/infra_tooling_account/infra_tooling_alerting_stack.py
vdektiarev: Didn't know we can do this, honestly 👍 Will rebase on latest changes and do that.

src/lib/event_mapper/impl/general_aws_event_mapper.py
vdektiarev: Agree. Created a card in Trello.

src/lib/event_mapper/aws_event_mapper.py
```diff
@@ -0,0 +1,40 @@
...
+    GeneralAwsEventMapper,
+    GlueEventMapper,
+    StepFunctionsEventMapper,
+    EventParsingException,
+)
+from ..settings import Settings
+
+
+class AwsEventMapper:
+    """Class designed to map AWS events to notification messages"""
```
npaskrobka, line 11: Not for this PR, but for the future let's make the Registry design pattern in a more common way across different components (notification service and alerting parser). 
In notification service we have it a bit differently - 2 methods: register & get (which returns a class object, not a fuction result). 
I think it is a good point for our next brainshtorming :)

src/lib/event_mapper/impl/__init__.py
```diff
@@ -0,0 +1,4 @@
+from .general_aws_event_mapper import GeneralAwsEventMapper
```
npaskrobka, line 1: A bit confused with the folder namings - impl :)
Would propose to rename folders impl -> event_mapper (since actual event_mappers are there), and event_mapper -> smth describing product component (like event_service, event_component, etc.). I think it can correlate whith the notification_service structure.

src/lib/event_mapper/aws_event_mapper.py
vdektiarev: Yeah, the intention here was to have a single entrypoint for mapping all types of AWS events, so that the lambda wouldn't know anything about the internal mapping logic/providers/registers/etc., but would just call _mapper.to_notification_messages(event)_ instead. And inside, we would have this _AwsEventMapper_ on top of _event_mapper_ package, which would do the event analysis and choosing the appropriate mapper from the _.impl_ (implementation) package, and pass the event there. And there we would have a general mapper, which contains common logic, and specific per AWS service mappers which do their specific parts of the process.

I agree that naming could be a bit ambiguous here (_aws_event_mapper_, _general_aws_event_mapper_), that's why I created an _.impl_ package to mark these modules non-usable directly in the library client code. I just didn't really want lambda to have to call additional _get_mapper_ function to get the appropriate mapper for the event. "Just do the mapping job, aws event mapper!" - that's the intention I wanted lambdas to have :) 

Please advise if we should stick fully to the approach in notification lambda (I agree that concepts alignment for similar solutions is important and makes the whole code a lot more maintainable), or do the renaming part only for now, or should we proceed just like that

PR #25, opened 2023-12-12
Feature/extract metrics glue

natmasslove:
+ implemented Extract Metrics Orchestration Lambda
  runs every 5 minutes
  invokes Extract Metrics Lambda for each monitoring group in our config
+ implemented Extract Metrics lambda
  it iterates through all resources listed in specific monitoring group
  for each resource it extracts metric information (data since the last update) and saves to timestream DB
  runs for all services, but actual extraction is implemented for Glue
  for other service - there are stubs currently, need to be implemented

* updated Metrics Lambda IAM role permissions (we need KMS key permissions while querying Timestream)
* changed naming convention for timestream tables (they no longer contain project and stage name - it's already in timestream DB name)
* GlueManager now operates using pydantic models (data classes)


src/lib/aws/aws_common_resources.py
```diff
@@ -0,0 +1,6 @@
+
+
+class AWS_Common_Resources:
```
nalkhimo, line 3: The class should be named in CamelCase manner, without underscores.

src/lib/aws/aws_common_resources.py
```diff
@@ -0,0 +1,6 @@
...
+    @classmethod
+    def get_Lambda_Powertools_Layer_Arn(cls, region):
```
nalkhimo, line 5: The method should be named in lowercase.

src/lib/aws/glue_manager.py
```diff
@@ -17,4 +59,19 @@ def get_all_job_names(self):
         except Exception as e:
             error_message = f"Error getting list of glue jobs : {e}"
             raise GlueManagerException(error_message)            
+        
+    def get_job_runs(self, job_name: str, since_time: datetime) -> list[JobRun]:
+        try:
+            outp = []
```
nalkhimo, line 65: This row is excess. outp will be set to an empty list below if the condition "x.StartedOn > since_time" is False.

src/lib/metrics_extractor/base_metrics_extractor.py
```diff
@@ -0,0 +1,48 @@
...
+        """
+        Extract metrics data from AWS and convert them to Timestream records.
+        Returns:
+            records (list): List of Timestream records
+            common_attributes (dict): Common attributes for all records (e.g. dimensions)
+        """
+        pass
+        
+
+    def write_metrics(self, records, common_attributes, timestream_table_writer: TimestreamTableWriter):
```
nalkhimo, line 46: Method "write_" should only write and not return anything.

src/lib/aws/timestream_manager.py
```diff
@@ -222,3 +238,60 @@ def get_MagneticStoreRetentionPeriodInDays(self):
...
+            response = self.timestream_query_client.query(QueryString=query)
+            first_record_data = response["Rows"][0]["Data"]
+            first_field_data = first_record_data[0]
+            if "ScalarValue" in first_field_data:
+                return first_field_data["ScalarValue"]
+            else:
+                return None
+        except Exception as e:
+            error_message = f"Error running query: {e}"
+            raise (TimestreamQueryException(error_message))
```
nalkhimo, line 271: I believe you do not need additional brackets here.

nalkhimo (approved):

src/lib/aws/aws_common_resources.py
natmasslove: fixed

src/lib/aws/aws_common_resources.py
natmasslove: fixed

src/lib/aws/glue_manager.py
natmasslove: fixed

src/lib/metrics_extractor/base_metrics_extractor.py
natmasslove: fixed

src/lib/aws/timestream_manager.py
natmasslove: fixed

npaskrobka (approved):
Great job! And nice pydentic implementation ;)

PR #27, opened 2023-12-15
Feature/settings wildcards

npaskrobka:
- Added wildcards replacements
- Added optional attribute for Settings class iam_role_extract_metrics
- Switched methods list_monitoring_groups() and get_monitoring_groups() to read from RAW settings
- Added lambda and step_functions managers
- Fixed settings bugs - in S3Manager add AccessDenied key to exception handler (to work from Lambda), moved replacements.json

Next steps:
- raw -> pre-processed (+ replacements.json) -> processed (+ defaults + wildcards) -- approach + implementation


infra_monitored_account/infra_monitored_account/infra_monitored_stack.py
```diff
@@ -89,48 +97,83 @@ def create_event_rules(self, cross_account_bus_role, cross_account_event_bus_arn
...
+        lambda_inline_policy = iam.Policy(
+            self,
+            "lambda-extract",
+            policy_name=AWSNaming.IAMPolicy(self, "lambda-extract"),
         )
-        lambda_inline_policy = iam.Policy(self, "lambda-extract", policy_name=AWSNaming.IAMPolicy(self,"lambda-extract"))
         lambda_inline_policy.add_statements(lambda_policy_statement)
 
+        # Step Functions Policy
+        states_policy_statement = iam.PolicyStatement(
```
natmasslove, line 157: Maybe, it's better to name it step_functions_policy_statement, step_functions_policy?

src/lambda_extract_metrics.py
```diff
@@ -38,14 +38,14 @@ def lambda_handler(event, context):
...
-    # doesn't proceed    
+    # doesn't proceed
 
     settings_s3_path = os.environ["SETTINGS_S3_PATH"]
     iam_role_name = os.environ["IAMROLE_MONITORED_ACC_EXTRACT_METRICS"]
     timestream_metrics_db_name = os.environ["TIMESTREAM_METRICS_DB_NAME"]
     monitoring_group_name = event.get("monitoring_group")
 
-    settings = Settings.from_s3_path(settings_s3_path)
+    settings = Settings.from_s3_path(settings_s3_path, iam_role_name)
```
natmasslove, line 48: I'd prefer to pass iam_role as named parameter.
Otherwise, it looks like IAM role is used to READ settings (while actually it is used to TRANSFORM they).

src/lib/settings/settings.py
```diff
@@ -287,10 +414,11 @@ def from_file_path(cls, base_path: str):
             monitoring_settings,
             recipients_settings,
             replacements_settings,
+            iam_role_extract_metrics,
         )
 
     @classmethod
-    def from_s3_path(cls, base_path: str):
+    def from_s3_path(cls, base_path: str, iam_role_extract_metrics: str = None):
```
natmasslove, line 421: Maybe we can have another name here, because inside the Settings class we don't extract metrics :)
rather it's a role to access and list monitored resources in general.

From my perspective, it might become misleading

src/lambda_test_settings.py
```diff
@@ -31,7 +22,7 @@ def lambda_handler(event, context):
     )
 
     # Lambda methods
-    account_id, region = "123456789", "eu-central-1"
+    account_id, region = "405389362913", "eu-central-1"
```
natmasslove, line 25: let's not put actual acc number here

src/lib/aws/s3_manager.py
```diff
@@ -52,7 +52,7 @@ def read_file(self, s3_path: str) -> str:
             settings_data = response["Body"].read().decode("utf-8")
             return settings_data
         except ClientError as e:
-            if e.response["Error"]["Code"] == "NoSuchKey":
+            if e.response["Error"]["Code"] in ["NoSuchKey", "AccessDenied"]:
```
natmasslove, line 55: Not sure if we need to handle File doesn't exist and Access Denied with the same message

src/lib/settings/settings.py
```diff
@@ -137,8 +171,93 @@ def _get_default_metrics_extractor_role_arn(self, account_id: str) -> str:
         return f"arn:aws:iam::{account_id}:role/role-salmon-cross-account-extract-metrics-dev"
 
     def _process_monitoring_groups(self):
-        # TODO: add wildcards replacement for all the resource types (glue, lambda, etc.)
-        pass
+        # Get resource names dict
+        resource_names = self._get_all_resource_names()
```
nalkhimo, line 175: Looks like you can move resource_names retrieving into  _replace_wildcards and pass into that method only m_env, m_res.

src/lib/settings/settings.py
```diff
@@ -137,8 +171,93 @@ def _get_default_metrics_extractor_role_arn(self, account_id: str) -> str:
...
+                "monitored_env_name_N": ["job1", ...,  "jobN"]]
+                },
+            "glue_workflows": {...},
+            ...
+            }"""
+        # Get all monitored accounts
+        monitored_accounts = self.general.get("monitored_environments", [])
+
+        # Initialize an empty dictionary for each resource type
+        resource_names = dict()
```
nalkhimo, line 199: There is a great container called collections.defaultdict which helps to create dict with no initial definition. You can take a look)

nalkhimo (approved):

src/lib/aws/s3_manager.py
npaskrobka: Good notice to make different messages, will do!

npaskrobka: Just want to leave also a little background for this change :) 
We are getting AccessDenied error when running code in Lambda, and NoSuchKey when debugging code locally, that's why both these keys should throw the same FileNotFound exception.

src/lib/settings/settings.py
npaskrobka: Good notice, but in such a case we would need to read glue jobs, lambdas for an account in each monitoring group (which can be 10 groups retrieving the same list of jobs). That's why preparing dictionary only once in the beginning

npaskrobka: By the way noticed misleading names - m_env should be m_grp, will change :)

infra_monitored_account/infra_monitored_account/infra_monitored_stack.py
npaskrobka: Done

src/lib/settings/settings.py
npaskrobka: Done

src/lambda_test_settings.py
npaskrobka: Done

src/lib/aws/s3_manager.py
npaskrobka: Done

src/lib/settings/settings.py
npaskrobka: Done, thanks for the great proposal! :)

src/lambda_extract_metrics.py
npaskrobka: Done

natmasslove (approved):

PR #30, opened 2023-12-20
Metrics collection for Glue Workflows and Step Functions

natmasslove:
+ Implemented metrics collection for Glue Workflows
+ Implemented metrics collection for Step Functions
* Fixed logging in extract metrics lambda (print -> logger.info)


src/lib/aws/step_functions_manager.py
```diff
@@ -1,4 +1,34 @@
 import boto3
+from datetime import datetime
+
+from pydantic import BaseModel
```
aryna-yukhnevich, line 4: not sure whether it's relevant to this PR, but should we add pydantic to the requirements.txt? :)

natmasslove: Actually, it's included in https://github.com/Soname-Solutions/salmon/blob/dev/src/requirements.txt already and, I guess, in that regard we should be fine.

But another problem linked to this, is that we have multiple environments (1 for src, 1 for infra tooling, 1 for infra monitored) and, I believe, we should reconsider this and have just one for the whole project.
This should be a tech-debt task for 1) refactoring 2) reflecting this in our deployment instructions

vdektiarev (approved):

PR #31, opened 2023-12-21
Replace bat scripts and env variables with context usage

vdektiarev:
I suggest using the following approach instead of .bat or .sh scripts usage:

`cdk deploy --all --profile <AWS_PROFILE> --context stage-name=<STAGE_NAME>
`
That would be more flexible and usable compared to env variables and .bat scripts usage.

Default STAGE_NAME=default will be used and logged if no stage is set.


infra_monitored_account/app.py
```diff
@@ -10,25 +11,31 @@
...
 
 settings = Settings.from_file_path("../config/settings")
 settings_validator.validate(settings)
 
 app = cdk.App()
+
+STAGE_NAME = app.node.try_get_context("stage-name")
+if STAGE_NAME is None:
+    logging.info('stage-name is not set. Using "default".')
+    STAGE_NAME = "default"
```
natmasslove, line 25: I'd rather throw an error instead of setting a default stage_name

natmasslove:
cdk deploy --all --profile <AWS_PROFILE> --context stage-name=<STAGE_NAME>

1. Agree with --context stage-name=<STAGE_NAME> - it's better that through a env var
2.  For  --profile <AWS_PROFILE> - I'd leave some flexibility for users:
it's either your command or 2nd option (which I prefer)
export AWS_DEFAULT_PROFILE=my_profile
cdk deploy --all --context stage-name=<STAGE_NAME>

vdektiarev:
> cdk deploy --all --profile <AWS_PROFILE> --context stage-name=<STAGE_NAME>
> 
> 1. Agree with --context stage-name=<STAGE_NAME> - it's better that through a env var
> 2. For  --profile <AWS_PROFILE> - I'd leave some flexibility for users:
>    it's either your command or 2nd option (which I prefer)
>    export AWS_DEFAULT_PROFILE=my_profile
>    cdk deploy --all --context stage-name=<STAGE_NAME>

Sure, it's possible, both ways would work.

For CDK context, you can even put this variable into cdk.json so that you wouldn't have to specify it manually on each run.

infra_monitored_account/app.py
```diff
@@ -21,8 +21,7 @@
 
 STAGE_NAME = app.node.try_get_context("stage-name")
 if STAGE_NAME is None:
-    logging.info('stage-name is not set. Using "default".')
-    STAGE_NAME = "default"
+    raise KeyError("stage-name conext variable is not set.")
```
natmasslove, line 24: Looks good. Let's just fix a typo "conext"

natmasslove (approved):

PR #33, opened 2023-12-27
add grafana stack

aryna-yukhnevich:
- optional Grafana stack added (if mandatory configs vpc id and security group id are not provided - this stack will be skipped)
- also some optional configs can be set: grafana_key_pair_name (if not provided, will be created in the Grafana stack), grafana_image (Bitnami Grafana image. Defaults to "bitnami-grafana-10.2.2-1-r02-linux-debian-11-x86_64-hvm-ebs-nami"), grafana_instance_type (Grafana Instance type. Defaults to "t3.micro")
- sample dashboard and user data script added to a separate folder infra_tooling_account/grafana
- Grafana provisioned config files will be generated considering each Service (Glue, Lambda, etc), uploaded to S3 settings bucket and copied to the Grafana instance at launch time. As a result, for each Service, a separate dashboard and data source will be provisioned in Grafana
- after the Stack deployment, to sign in, go to http://<grafana-instance-public-ip>:3000 and login under admin user and password taken from the corresponding Grafana secret created in SM.

### DIFF
--- a/config/sample_settings/general.json
+++ b/config/sample_settings/general.json
@@ -3,7 +3,9 @@
         "name": "Tooling Account [<<env>>]",
         "account_id": "<<tooling_account_id>>",
         "region": "eu-central-1",
-        "metrics_collection_interval_min" : 5
+        "metrics_collection_interval_min" : 5,
+        "grafana_vpc_id": "<<vpc_id>>", 
+        "grafana_security_group_id": "<<security_group_id>>"
     },
     "monitored_environments": [
         {
--- a/infra_tooling_account/app.py
+++ b/infra_tooling_account/app.py
@@ -10,6 +10,7 @@ from infra_tooling_account.infra_tooling_alerting_stack import InfraToolingAlert
 from infra_tooling_account.infra_tooling_monitoring_stack import (
     InfraToolingMonitoringStack,
 )
+from infra_tooling_account.infra_tooling_grafana_stack import InfraToolingGrafanaStack
 
 from lib.settings.settings import Settings
 from lib.settings.cdk import settings_validator
@@ -27,7 +28,6 @@ TAGS = {"project_name": PROJECT_NAME, "stage_name": STAGE_NAME}
 
 settings = Settings.from_file_path("../config/settings")
 settings_validator.validate(settings)
-
 
 app = cdk.App()
 common_stack = InfraToolingCommonStack(
@@ -57,5 +57,20 @@ monitoring_stack = InfraToolingMonitoringStack(
 
 alerting_stack.add_dependency(common_stack)
 monitoring_stack.add_dependency(common_stack)
+
+if all(setting is not None for setting in settings.get_grafana_mandatory_settings()):
+    grafana_stack = InfraToolingGrafanaStack(
+        app,
+        f"cf-{PROJECT_NAME}-InfraToolingGrafanaStack-{STAGE_NAME}",
+        tags=TAGS,
+        stage_name=STAGE_NAME,
+        project_name=PROJECT_NAME,
+        settings=settings,
+        env={
+            "region": os.getenv('CDK_DEFAULT_REGION'),
+            "account":os.getenv('CDK_DEFAULT_ACCOUNT'),
+        },
+    )
+    grafana_stack.add_dependency(common_stack)
 
 app.synth()
--- a/infra_tooling_account/infra_tooling_account/infra_tooling_grafana_stack.py
+++ b/infra_tooling_account/infra_tooling_account/infra_tooling_grafana_stack.py
@@ -1,0 +1,386 @@
+import yaml
+import json
+from aws_cdk import (
+    Stack,
+    CfnOutput,
+    Fn,
+    aws_s3 as s3,
+    aws_iam as iam,
+    aws_ec2 as ec2,
+    aws_s3_deployment as s3deploy,
+    aws_secretsmanager as secretsmanager,
+)
+from constructs import Construct
+from lib.aws.aws_naming import AWSNaming
+from lib.settings.settings import Settings
+from lib.core.constants import SettingConfigs
+from lib.core.grafana_config_generator import (
+    generate_dashboard_model,
+    generate_datasources_config_section,
+    generate_dashboards_config_section,
+    generate_user_data_script,
+)
+
+
+class InfraToolingGrafanaStack(Stack):
+    """
+    This class represents a stack for Grafana instance in AWS CloudFormation.
+
+    Attributes:
+        stage_name (str): The stage name of the deployment, used for naming resources.
+        project_name (str): The name of the project, used for naming resources.
+
+    """
+
+    def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
+        """
+        Initialize the GrafanaStack.
+
+        Args:
+            scope (Construct): The CDK app or stack that this stack is a part of.
+            construct_id (str): The identifier for this stack.
+            **kwargs: Arbitrary keyword arguments. Specifically looks for:
+                - project_name (str): The name of the project. Used for naming resources. Defaults to None.
+                - stage_name (str): The name of the deployment stage. Used for naming resources. Defaults to None.
+                - settings (Tuple): Tuple containing settings. Defaults to None.
+
+        """
+        self.stage_name = kwargs.pop("stage_name", None)
+        self.project_name = kwargs.pop("project_name", None)
+        self.settings: Settings = kwargs.pop("settings", None)
+
+        super().__init__(scope, construct_id, **kwargs)
+
+        (
+            input_timestream_database_arn,
+            input_timestream_kms_key_arn,
+            input_settings_bucket_arn,
+            settings_bucket,
+        ) = self.get_common_stack_references()
+
+        (
+            grafana_vpc_id,
+            grafana_security_group_id,
+        ) = self.settings.get_grafana_mandatory_settings()
+
+        (
+            grafana_key_pair_name,
+            grafana_bitnami_image,
+            grafana_instance_type,
+        ) = self.settings.get_grafana_optional_settings()
+
+        (
+            grafana_admin_secret_name,
+            grafana_admin_secret_arn,
+        ) = self.create_grafana_admin_secret()
+
+        # Create Grafana key pair if not provided
+        if not grafana_key_pair_name:
+            grafana_key_pair_name = self.create_grafana_key_pair()
+
+        self.prepare_and_upload_grafana_configuration_files(settings_bucket)
+
+        grafana_instance = self.create_grafana_instance(
+            grafana_vpc_id=grafana_vpc_id,
+            grafana_security_group_id=grafana_security_group_id,
+            input_timestream_kms_key_arn=input_timestream_kms_key_arn,
+            input_timestream_database_arn=input_timestream_database_arn,
+            input_settings_bucket_arn=input_settings_bucket_arn,
+            grafana_admin_secret_name=grafana_admin_secret_name,
+            grafana_admin_secret_arn=grafana_admin_secret_arn,
+            grafana_bitnami_image=grafana_bitnami_image,
+            grafana_key_pair_name=grafana_key_pair_name,
+            grafana_instance_type=grafana_instance_type,
+            settings_bucket_name=settings_bucket.bucket_name,
+        )
+
+        output_grafana_instance_public_ip = CfnOutput(
+            self,
+            "GrafanaPublicIp",
+            value=grafana_instance.instance_public_ip,
+            description="The Public IP of the Grafana Instance",
+            # To sign in to Grafana, go to http://<grafana-instance-public-ip>:3000
+            export_name=AWSNaming.CfnOutput(self, "grafana-instance-public-ip"),
+        )
+
+    def get_common_stack_references(self) -> tuple[str, str, str, s3.Bucket]:
+        """
+        Retrieves common stack references required for the stack's operations.
+        These include the Timestream database ARN, Timestream KMS key ARN, settings S3 bucket and its ARN.
+
+        Returns:
+            tuple: A tuple containing references to the Timestream database ARN, Timestream KMS key ARN,
+                   settings S3 bucket and its ARN.
+        """
+
+        input_timestream_database_arn = Fn.import_value(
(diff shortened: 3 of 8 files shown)


infra_tooling_account/infra_tooling_account/infra_tooling_grafana_stack.py
```diff
@@ -0,0 +1,386 @@
...
+            grafana_instance_type=grafana_instance_type,
+            settings_bucket_name=settings_bucket.bucket_name,
+        )
+
+        output_grafana_instance_public_ip = CfnOutput(
+            self,
+            "GrafanaPublicIp",
+            value=grafana_instance.instance_public_ip,
+            description="The Public IP of the Grafana Instance",
+            # To sign in to Grafana, go to http://<grafana-instance-public-ip>:3000
```
natmasslove, line 102: I guess, it would be even better to have Grafana URL as output (so user can just copy and paste it).
Ot maybe, expose both IP and URL from stack outputs

config/sample_settings/general.json
```diff
@@ -3,7 +3,9 @@
         "name": "Tooling Account [<<env>>]",
         "account_id": "<<tooling_account_id>>",
         "region": "eu-central-1",
-        "metrics_collection_interval_min" : 5
+        "metrics_collection_interval_min" : 5,
+        "grafana_vpc_id": "<<vpc_id>>",
```
natmasslove, line 7: For me, it looks a bit more logical to have a separate section for Grafana:
"grafana_instance" : {
   "vpc_id" :
   "security_group_id" :
}
If we "grafana_instance" is present is general, then we create a stack. And we don't otherwise.

src/lib/settings/settings.py
```diff
@@ -280,6 +280,32 @@ def get_metrics_collection_interval_min(self) -> int:
...
+        """
+        Get optional grafana settings:
+            - grafana_key_pair_name: Key pair name. if not provided, will be created in the Grafana stack
+            - grafana_image: Bitnami Grafana image. Defaults to "bitnami-grafana-10.2.2-1-r02-linux-debian-11-x86_64-hvm-ebs-nami"
+            - grafana_instance_type: Grafana Instance type. Defaults to "t3.micro"
+        """
+        grafana_settings = self.processed_settings[SettingFileNames.GENERAL]["tooling_environment"]
+        return (
+            grafana_settings.get("grafana_key_pair_name"),
+            grafana_settings.get("grafana_bitnami_image", "bitnami-grafana-10.2.2-1-r02-linux-debian-11-x86_64-hvm-ebs-nami"),
```
natmasslove, line 306: both default values for Image and Instance size - I'd prefer to have them in constants.py

src/lib/settings/settings.py
```diff
@@ -280,6 +280,32 @@ def get_metrics_collection_interval_min(self) -> int:
...
+            - grafana_vpc_id: VPC ID for Grafana instance
+            - grafana_security_group_id: Security Group ID for Grafana instance
+        If not provided, the Grafana stack will be skipped.
+        """
+        grafana_settings = self.processed_settings[SettingFileNames.GENERAL]["tooling_environment"]
+        return (
+            grafana_settings.get("grafana_vpc_id"),
+            grafana_settings.get("grafana_security_group_id"))
+
+    def get_grafana_optional_settings(self) -> tuple[str, str, str]:
```
natmasslove, line 296: I'm not sure we need to separate mandatory and optional settings into a different methods.
Maybe, we can just have get_grafana_settings?

src/lib/settings/settings.py
```diff
@@ -280,6 +280,32 @@ def get_metrics_collection_interval_min(self) -> int:
...
+        """
+        Get optional grafana settings:
+            - grafana_key_pair_name: Key pair name. if not provided, will be created in the Grafana stack
+            - grafana_image: Bitnami Grafana image. Defaults to "bitnami-grafana-10.2.2-1-r02-linux-debian-11-x86_64-hvm-ebs-nami"
+            - grafana_instance_type: Grafana Instance type. Defaults to "t3.micro"
+        """
+        grafana_settings = self.processed_settings[SettingFileNames.GENERAL]["tooling_environment"]
```
natmasslove, line 303: filling in default values, preferably, should be done when preparing processing settings (raw->processing)

src/lib/aws/aws_naming.py
```diff
@@ -79,6 +79,16 @@ def TimestreamTable(cls, stack_obj: object, meaning: str) -> str:
...
         outp = f"{prefix}-{meaning}" # Table lives inside DB, so we identify project and stage names by DB
         return outp
+        
+    @classmethod
+    def EC2(cls, stack_obj: object, meaning: str) -> str:
+        prefix = "ec2"
+        return AWSNaming.__resource_name_with_check(stack_obj, prefix, meaning)
+    
+    @classmethod
+    def SM(cls, stack_obj: object, meaning: str) -> str:
```
natmasslove, line 89: Let's call it "secret" :)
For me, it required to search through code to understand what does SM mean :)

infra_tooling_account/app.py
```diff
@@ -58,4 +58,19 @@
...
+if all(setting is not None for setting in settings.get_grafana_mandatory_settings()):
+    grafana_stack = InfraToolingGrafanaStack(
+        app,
+        f"cf-{PROJECT_NAME}-InfraToolingGrafanaStack-{STAGE_NAME}",
+        tags=TAGS,
+        stage_name=STAGE_NAME,
+        project_name=PROJECT_NAME,
+        settings=settings,
+        env={
+            "region": os.getenv('CDK_DEFAULT_REGION'),
```
natmasslove, line 70: If I'm not wrong - this requires CDK_DEFAULT_REGION env var to be set, right?

Maybe, then, 
current_region = Stack.of(self).region
looks as a more universal approach, as we let CDK to decide how to get current region.

infra_tooling_account/infra_tooling_account/infra_tooling_grafana_stack.py
aryna-yukhnevich: thank you! agree, will adjust accordingly

config/sample_settings/general.json
aryna-yukhnevich: yes, good idea :)

src/lib/aws/aws_naming.py
aryna-yukhnevich: sure :)

npaskrobka (commented):
Really great job!
Regarding settings part just one small comment, for the rest totally agree with @natmasslove  - optional and mandatory settings can be combined, and defaults should probably be moved to the cached property processed_settings (to be in one place along with the other defaults)

infra_tooling_account/infra_tooling_account/infra_tooling_grafana_stack.py
aryna-yukhnevich: Grafana URL is in the output now. I think only URL would be enough, but let me know if any further adjustments are required

config/sample_settings/general.json
aryna-yukhnevich: A separate section added

src/lib/settings/settings.py
aryna-yukhnevich: added to constants.py

src/lib/settings/settings.py
aryna-yukhnevich: combined into one method

src/lib/aws/aws_naming.py
aryna-yukhnevich: updated accordingly

infra_tooling_account/app.py
aryna-yukhnevich: unfortunately, Stack.of(self).region does not work for me, I'll try to investigate further on this

src/lib/settings/settings.py
```diff
@@ -280,6 +280,32 @@ def get_metrics_collection_interval_min(self) -> int:
...
         ]
+    
+    def get_grafana_mandatory_settings(self) -> tuple[str, str]:
+        """
+        Get mandatory grafana settings:
+            - grafana_vpc_id: VPC ID for Grafana instance
+            - grafana_security_group_id: Security Group ID for Grafana instance
+        If not provided, the Grafana stack will be skipped.
+        """
+        grafana_settings = self.processed_settings[SettingFileNames.GENERAL]["tooling_environment"]
```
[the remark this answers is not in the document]
aryna-yukhnevich: updated, feel free to review and let me know if any further adjustments still required

infra_tooling_account/infra_tooling_account/infra_tooling_grafana_stack.py
```diff
@@ -0,0 +1,383 @@
...
+        Generates and uploads Grafana configuration files to S3 bucket.
+        For each Service, a separate dashboard and data source will be provisioned.
+        The files will be uploaded to settings/grafana dedicated folder.
+
+        Args:
+            settings_bucket (s3.Bucket): Settings S3 Bucket
+        """
+        metric_table_names = {x: f"{x}-metrics" for x in SettingConfigs.RESOURCE_TYPES}
+        services = metric_table_names.keys()
+        timestream_database_name = AWSNaming.TimestreamDB(
```
vdektiarev, line 184: Can we use the output from infra_tooling_common_stack? There is an output with the DB name from it.

infra_tooling_account/infra_tooling_account/infra_tooling_grafana_stack.py
```diff
@@ -0,0 +1,383 @@
...
+        grafana_role.add_to_policy(
+            # to be able to make Timestream API calls from Grafana instance
+            iam.PolicyStatement(
+                actions=[
+                    "timestream:DescribeEndpoints",
+                    "timestream:ListDatabases",
+                    "timestream:SelectValues",
+                ],
+                effect=iam.Effect.ALLOW,
+                resources=["*"],
```
vdektiarev, line 316: Is it possible to limit the permissions to our specific Timestream DB only?

natmasslove: I believe, for DescribeEndpoints action, we should keep "*" - kind of internal limitation from Timestream.
For others - I'm not sure, let's check and tighten permissions is possible

infra_tooling_account/infra_tooling_account/infra_tooling_grafana_stack.py
aryna-yukhnevich: unfortunately, if I use the output with the DB name in the Grafana generated config files that will be uploaded to S3 bucket, I'm getting this error:
RuntimeError: Invalid CloudFormation reference. Key must start with any of "Ref" or "Fn::GetAtt" or "Fn::Select". Got {"Fn::ImportValue":"output-salmon-metrics-events-db-name-devay"}

will try to research further on this

infra_tooling_account/infra_tooling_account/infra_tooling_grafana_stack.py
aryna-yukhnevich: Yes, right. When I set the DescribeEndpoints action for my Timstream DB only, I'm getting this error in Grafana:

AccessDeniedException: User: arn:aws:sts::account_id:assumed-role/role-salmon-grafana-instance-devay/i-0a55a4093dc8364e7 is not authorized to perform: timestream:DescribeEndpoints because no identity-based policy allows the timestream:DescribeEndpoints action

aryna-yukhnevich: For others - I'll re-check, thank you!

aryna-yukhnevich: the same error "AccessDeniedException" if the action timestream:SelectValues is set only for my DB
and as for "timestream:ListDatabases", as I can see, the Timestream plugin will not work for new data sources added in Grafana UI, since we will not be able to select the database name while establishing the connection

natmasslove: I'm fine with leaving "*" if it's needed, but, I guess there might be more details :), so let's have a separate discussion about this during the next sync-up

infra_tooling_account/app.py
natmasslove: what is the error you are getting? Let's have a follow-up discussion on Jan, 3rd :)

npaskrobka (approved):

infra_tooling_account/app.py
aryna-yukhnevich: updated as discussed

config/sample_settings/README.md
```diff
@@ -11,6 +11,17 @@ Contains the following parts:
 
 - metrics_collection_interval_min - interval (in minutes) for extracting metrics from monitored environments
 
+#### Grafana settings
+The Grafana stack will be deployed only if the Grafana related settings are provided in the "grafana_instance" section, nested within the "tooling_environment" configuration. \
+The deployment requires two mandatory configurations: "grafana_vpc_id" and "grafana_security_group_id":
+- "grafana_vpc_id": the ID of the Amazon VPC where the Grafana instance will be deployed.
+- "grafana_security_group_id": the ID of the security group that will be associated with the Grafana instance.
```
natmasslove, line 18: maybe, it makes sense to describe security group setup: like, you should have port #<x> opened etc.
Also, for VPC - should have at least 1 public subnet (or something like this)

config/sample_settings/README.md
```diff
@@ -11,6 +11,17 @@ Contains the following parts:
...
+
+ Additionally, several optional configurations are available to customize the Grafana deployment:
+- "grafana_key_pair_name": the name of the key pair to be associated with the Grafana instance. If not provided, a new key pair will be created during the stack deployment.
+- "grafana_bitnami_image": the Bitnami Grafana image from AWS Marketplace.
```
natmasslove, line 22: here, also, let's describe default values

natmasslove (approved):

config/sample_settings/README.md
aryna-yukhnevich: updated accordingly, thank you!

config/sample_settings/README.md
aryna-yukhnevich: updated accordingly, thank you!

PR #38, opened 2024-01-09
Feature/timestream optimize

natmasslove:
+ Optimized getting metrics last update time for resources:
Now it retrieves last_update_time for each resource in orchestration lambda and passes if as a payload to each extractor lamba.
Extractor lambda relies on these parameter, but still able to fallback to older mechanism if no parameter is given

* unified data structure in metrics tables:

column now is "resource_name" in all tables (instead of "job_name", "step_function_name",...)
metric_name now is "execution" (instead of "job_execution", ...)
* get since_time method moved to base class (as the only difference was queried field_names for each resource_type)
* renamed variables to reflect unified terminology (resource_type and resource_name)


aryna-yukhnevich (approved):

src/lib/core/time_utils.py
```diff
@@ -1,5 +1,6 @@
 from datetime import datetime
 import time
+import dateutil
```
npaskrobka, line 4: Wouldn't it be better to name this module datetime_utils? Seems that it works with datetime, not just time :)

src/lambda_extract_metrics.py
```diff
@@ -8,7 +8,7 @@
 from lib.settings import Settings
 from lib.core.constants import SettingConfigs
 
-from lib.metrics_extractor import MetricsExtractorProvider
+from lib.metrics_extractor import MetricsExtractorProvider, get_last_update_time
```
npaskrobka, line 11: I would not import the functions like classes, because it is harder to find where this function comes from (for classes it is obvious since file name is the same as class name). 

```
from lib.metrics_extractor import MetricsExtractorProvider
from lib.metrics_extractor.metrics_extractor_utils import get_last_update_time
```

src/lib/metrics_extractor/base_metrics_extractor.py
```diff
@@ -1,36 +1,49 @@
...
+
+        # check if table is empty
+        query = f'SELECT count(*) FROM "{self.timestream_db_name}"."{self.timestream_metrics_table_name}"'
+        count_rec = int(queryRunner.execute_scalar_query(query=query))
+        if count_rec == 0:
+            return None
+
+        query = f'SELECT max(time) FROM "{self.timestream_db_name}"."{self.timestream_metrics_table_name}" WHERE {self.RESOURCE_NAME_COLUMN_NAME} = \'{self.resource_name}\''
+        last_date = queryRunner.execute_scalar_query_date_field(query=query)
+        return last_date
```
npaskrobka, line 46: A bit confused why we need this method also in BaseMetricsExtractor, seems only the one from metrics_extractor_utils is being used :)

src/lib/metrics_extractor/metrics_extractor_utils.py
```diff
@@ -0,0 +1,74 @@
...
+            query_cnt = (
+                f'SELECT count(*) FROM "{timestream_db_name}"."{timestream_table_name}"'
+            )
+            result_cnt = query_runner.execute_scalar_query(query_cnt)
+            if result_cnt == "0":
+                logger.info(f"No data in table {timestream_table_name}, skipping..")
+                output_dict[resource_type] = {}
+            else:
+                logger.info(e)
+                raise (e)
```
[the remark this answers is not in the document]
natmasslove: We are potentially getting exception from query_runner.execute_scalar_query(query_cnt), so it's already
a MetricsExtractorUtilsException

src/lib/core/time_utils.py
natmasslove: It's a python built-in module :) Can't rename :)

vdektiarev: Nastassia's probably meaning this very file she's reviewing

vdektiarev (approved):

src/lib/core/time_utils.py
natmasslove: Oh, sorry. Silly me :)
Fixed. Renamed to datetime_utils.py

src/lib/metrics_extractor/base_metrics_extractor.py
natmasslove: Actually, both are used in lambda_extract_metrics. 
And it was the idea - if we don't get last_update_times from orchestrator lambda (e.g. we ran extract lambda manually for debug, it takes last_update_time from timestream table directly, using this method):

                # 2. Get time of this entity's data latest update (we append data since that time only)
                since_time = get_last_update_time(
                    last_update_times, resource_type, resource_name
                )
                ...

                if since_time is None:
                    # if last_update_time was not given or missing - query for specific resource directly from table
...
                    since_time = metrics_extractor.get_last_update_time(
                        timestream_query_client=timestream_query_client
                    )

src/lambda_extract_metrics.py
natmasslove: fixed, thanks

src/lib/core/time_utils.py
npaskrobka: yes, meant exactly this, thanks! unfortunatelly cannot put comment on the file name :)

src/lib/metrics_extractor/base_metrics_extractor.py
npaskrobka: I see now, thanks! Yes, it is quite useful then, let's leave as it is

src/lib/metrics_extractor/metrics_extractor_utils.py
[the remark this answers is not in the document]
npaskrobka: Not really :) it will be TimeStreamQueryWriter exception, and query we run in several places in code.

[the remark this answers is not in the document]
npaskrobka: But probably let leave it till the story with the code refactoring, there it is exactly the task to unify the approaches for all the stuff :)
I'll add error handling topic there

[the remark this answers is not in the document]
natmasslove: Fixed :)

PR #39, opened 2024-01-09
Connect Grafana and CW Log Stream with alert events

aryna-yukhnevich:
- The dashboards samples based on the Timestream tables were separated for each Resource type accordingly;
- Small fix was applied to the Grafana secret so to exclude the character "-" from the Grafana password;
- Additionally the dashboard based on the CloudWatch stream with Alert Events was added;

image

and each message can be inspected:
image


infra_tooling_account/infra_tooling_account/grafana/glue_jobs_dashboard.sample.json
```diff
@@ -3,14 +3,11 @@
     "list": [
```
natmasslove, line 3: According to these file's usage it might be better to call them .template.json, not .sample.json?

src/lib/core/grafana_config_generator.py
```diff
@@ -1,73 +1,191 @@
...
+            for key, value in obj.items():
+                obj[key] = replace_recursive(value)
+        elif isinstance(obj, list):
+            for i, item in enumerate(obj):
+                obj[i] = replace_recursive(item)
+        elif isinstance(obj, str) and obj in replacements:
+            obj = replacements[obj]
+        return obj
+
+    return replace_recursive(json_data)
```
npaskrobka, line 33: We have the same replacements use-case and implementation in settings. It makes sense to have it only once. 
Can you please move this method in lib.core.json_utils, so that I can also use it in settings afterwards not to duplicate the code?
Thanks in advance!

aryna-yukhnevich: thank you for your comment, sure :)

infra_tooling_account/infra_tooling_account/grafana/glue_jobs_dashboard.sample.json
aryna-yukhnevich: yes, thank you :)

src/lib/core/grafana_config_generator.py
aryna-yukhnevich: updated accordingly

npaskrobka (approved):

natmasslove (approved):

PR #40, opened 2024-01-11
Feature/alert events subject

vdektiarev:


src/lib/event_mapper/resource_type_resolver.py
```diff
@@ -0,0 +1,32 @@
+from ..core.constants import SettingConfigResourceTypes as types
+
+
+class ResourceTypeResolver:
+    _source_detail_type_map = {
```
vdektiarev, line 5: I personally don't like that resolver structure and resolution algorithm, but please advise if that can be done more beautifully in Python without if-else structures 🙃

src/lambda_alerting.py
```diff
@@ -55,7 +56,7 @@ def write_event_to_cloudwatch(
     logged_event["event"] = event
     logged_event["monitored_environment"] = monitored_env_name
     logged_event["resource_name"] = resource_name
-    logged_event["service_name"] = service_name
+    logged_event["service_name"] = resource_type
```
natmasslove: let's rename a field to "resource_type" as well

vdektiarev: Done

src/lib/event_mapper/impl/step_functions_event_mapper.py
```diff
@@ -12,14 +12,18 @@ def get_resource_name(self, event):
...
+        return event["detail"]["status"]
 
     def get_event_status(self, event):
         return event["detail"]["status"]
 
     def get_event_severity(self, event):  # todo: implement
-        return "Unknown"
+        return (
+            EventSeverity.ERROR
+            if self.get_resource_state(event) in ("FAILED", "TIMED_OUT")
```
natmasslove: here I'd prefer to re-use failure stated definitions from step_functions_manager:

class StepFunctionsManager:
    STATES_SUCCESS = ["SUCCEEDED"]
    STATES_FAILURE = ["FAILED", "ABORTED", "TIMED_OUT"]
    
The  same is for glue jobs and glue_workflows (we also have such constants defined)

And, typically, there is a method there "is_final_state" - might be useful for us to decide if we want to put the event into cloudwatch log

src/lib/event_mapper/impl/glue_job_event_mapper.py
```diff
@@ -0,0 +1,35 @@
...
+    def get_resource_name(self, event):
+        return event["detail"]["jobName"]
+
+    def get_resource_state(self, event):
+        return event["detail"]["state"]
+
+    def get_event_severity(self, event):
+        return (
+            EventSeverity.ERROR
+            if self.get_resource_state(event) in ("FAILED", "TIMEOUT")
```
natmasslove: please refer to my comment regarding Step Functions states

src/lib/event_mapper/impl/glue_crawler_event_mapper.py
```diff
@@ -0,0 +1,36 @@
...
+    def __init__(self, settings: Settings):
+        super().__init__(settings)
+
+    def get_resource_name(self, event):
+        return event["detail"]["crawlerName"]
+
+    def get_resource_state(self, event):
+        return event["detail"]["state"]
+
+    def get_event_severity(self, event):
```
natmasslove: as per sync-up discussion, we might need event_result rather than event_severity

vdektiarev: Already fixed

src/lib/event_mapper/impl/step_functions_event_mapper.py
vdektiarev: Changed to managers usage to determine the event result, however I'm not sure it's really beneficial to add another public method to event mapper to get the info if the state is final, while still provisioning the event result to mapper clients because we have a logic which filters out both info and success event results - for alerts. What do you think about keeping the filtration logic about the "monitorable" events a bit more flexible like it's implemented now and have it aligned with the "alertable" events filtration logic?

src/lib/event_mapper/impl/step_functions_event_mapper.py
```diff
@@ -12,14 +13,18 @@ def get_resource_name(self, event):
         arn = event["detail"]["stateMachineArn"]
         return arn.split("stateMachine:")[1]
 
-    def get_service_name(self):
-        return SettingConfigResourceTypes.STEP_FUNCTIONS
+    def get_resource_state(self, event):
+        return event["detail"]["status"]
 
     def get_event_status(self, event):
         return event["detail"]["status"]
```
aryna-yukhnevich, line 20: looks like get_event_status needs to be removed here

aryna-yukhnevich (approved):

src/lambda_alerting.py
```diff
@@ -89,28 +56,47 @@ def lambda_handler(event, context):
...
+            event_result,
+            event,
+        )
+    else:
+        logger.info(f"Event result is not monitorable: {event_result}")
+
+    return {"messages": notification_messages}
 
 
 if __name__ == "__main__":
```
natmasslove, line 102: timestream db/table names seems obsolete in local debug section

natmasslove (approved):

PR #47, opened 2024-01-22
Feature/secret manager

nalkhimo:


src/lambda_test_notification_service.py
```diff
@@ -66,12 +57,20 @@ def lambda_handler(event, context):
 
     message = Message(formatted_message, message_subject)
 
+    secret_client = SecretManager(region_name="eu-north-1")
```
natmasslove, line 60: 1. Please don't use region other than eu-central-1
2. resources created in eu-north-1 and other regions - should be deleted
3. I guess, here we don't need parameter "region_name" - just read secret from default region

src/lib/aws/secret_manager.py
```diff
@@ -0,0 +1,45 @@
...
+    """Manages interactions with AWS Secrets Manager."""
+
+    def __init__(self, region_name: str, secret_client=None) -> None:
+        """Initiate class SecretManager.
+
+        Args:
+            secret_client: Boto3 SES client for AWS interactions.
+        """
+        self._secret_client = (
+            boto3.client(service_name="secretsmanager", region_name=region_name)
```
natmasslove, line 23: Here, also, I'm not sure we need to use region_name param here

nalkhimo: Done.

src/test_secret_manager.py
```diff
@@ -0,0 +1,7 @@
+from lib.aws.secret_manager import SecretManager
```
natmasslove, line 1: I guess, we don't need this file to be merged in dev branch

natmasslove (approved):
one comment - please remove the file which seems to be a temporary testing script

PR #48, opened 2024-01-22
Added lambda_metrics_extractor

npaskrobka:
- added lambda_metrics_extractor
- extended lambda policy in imfra_monitored_account
- "assume role" functionality moved to BaseMetricsExtractor (for lambda to create 2 managers - Lambda and CloudWatch)
- fixed small bug in settings get_monitoring_group_content() method

### DIFF
--- a/src/lambda_extract_metrics.py
+++ b/src/lambda_extract_metrics.py
@@ -4,7 +4,6 @@ import boto3
 import logging
 from itertools import groupby
 
-from lib.aws.sts_manager import StsManager
 from lib.aws.timestream_manager import TimestreamTableWriter
 from lib.aws.aws_naming import AWSNaming
 from lib.settings import Settings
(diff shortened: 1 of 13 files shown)


src/lambda_extract_metrics.py
```diff
@@ -58,7 +40,10 @@ def process_individual_resource(
     # 1. Create an extractor object for a specific service
     metrics_extractor = MetricsExtractorProvider.get_metrics_extractor(
         resource_type=resource_type,
-        aws_service_client=aws_service_client,
+        account_id=account_id,
```
natmasslove, line 43: I know we need 2 boto3 clients to get lambda metrics, but I'm thinking of ways other than passing 4 arguments all the way down thru layer.

Just an idea - let's consider, maybe, it can be a bit more estethically pleasing:

create lib.aws class Boto3ClientCreator
__init__ (account_id, region, role_name)
self.get_client(service_name)

and we can pass just client_creator var instead of all these.
Makes sense?

src/lib/aws/lambda_manager.py
```diff
@@ -1,4 +1,50 @@
...
+    def IsReportEvent(self) -> bool:
+        return LambdaManager.MESSAGE_PART_REPORT in self.message
+
+    @property
+    def IsErrorEvent(self) -> bool:
+        return LambdaManager.MESSAGE_PART_ERROR in self.message
+
+    @property
+    def Duration(self) -> float:
+        duration_match = re.search(r"Duration: ([0-9.]+) ms", self.message)
```
natmasslove, line 28: Nice :)
👍

natmasslove (approved):
no objections other than potential client_creator implementation

src/lambda_extract_metrics.py
npaskrobka: Thanks for the commet, totally makes sense, done :)

natmasslove (approved):

PR #56, opened 2024-02-07
Feature/glue workflow alerts

natmasslove:


natmasslove:
It's intended to have a follow-up task for event_mapper -> don't take "event" as a param for each and every method, but just take it from self.event.

Didn't want to add it into this PR to keep complexity relatively low

aryna-yukhnevich (approved):

src/lib/event_mapper/impl/general_aws_event_mapper.py
```diff
@@ -100,7 +111,7 @@ def create_table_row(self, values: list, style: str = None) -> dict:
             row["style"] = style
         return row
 
-    def to_message(self, monitored_env_name: str, event: dict) -> dict:
```
vdektiarev, line 103: We are now both assigning an event as a class field in the constructor, and accept an event in method arguments - maybe we should leave only one event in the class? Do we really need the event as the class field?

src/lib/event_mapper/impl/general_aws_event_mapper.py
```diff
@@ -17,6 +18,15 @@ class GeneralAwsEventMapper(ABC):
         to_notification_messages(dict): maps AWS event object to a list of notification message objects
     """
 
+    def __init__(
+        self,
+        event: dict,
+        settings: Settings
+    ):
+        self.event = event
+        self.settings = settings
```
vdektiarev, line 27: Not sure if we need Settings in the class fields. Looks like we use it only to determine the _monitored_env_name_

vdektiarev: What do you think about:
1. Assigning Settings only as a **class field**
2. Creating a method which calculates _monitored_env_name_, accepting the event as an argument. _general_aws_event_mapper_ will have the default implementation, while _glue_workflow_event_mapper_ will overrride it. The method will use "_settings_" class field.
3. Accepting only "_event_" as a mapping method argument, and, when _monitored_env_name_ is needed, the mapper will call the corresponding method.

vdektiarev (approved):

src/lib/event_mapper/impl/general_aws_event_mapper.py
vdektiarev: Decided to proceed with class fields approach for these classes to be consistent with others like senders, metrics extractors etc.

PR #57, opened 2024-02-12
Require SMTP credentials only when it's used

vdektiarev:


aryna-yukhnevich (approved):

src/lambda_notification.py
```diff
@@ -78,21 +78,27 @@ def lambda_handler(event, context):
...
-            message=message,
-            ses_sender=delivery_method.get("sender_email"),
-            recipients=delivery_options_info.get("recipients"),
-            smtp_sender=smtp_secret["SMTP_SENDER"],
-            smtp_server=smtp_secret["SMTP_SERVER"],
-            smtp_port=smtp_secret["SMTP_PORT"],
-            smtp_login=smtp_secret["SMTP_LOGIN"],
-            smtp_password=smtp_secret["SMTP_PASSWORD"],
-        )
+        if delivery_method_type == "SMTP":
```
natmasslove, line 81: on a second thought, can we consider here using SenderProvider?

I mean instead of multiple if-else for each type.
And, actually, it wouldn't be necessary to get secret's data here - it'd be a responsibility of sender (located by sender provider).

what do you think?

vdektiarev: Yeah, it's possible, just need to pass smtp secret as kwargs, I believe.
I was just not too confident about delegating secrets manager manipulation logic to Sender classes, seemed a bit unnatural for them to take additional responsibility, i.e. retrieving some creds from a separate AWS Service. But can do that to avoid these if-elses which I don't really like either.

natmasslove: Maybe we can consider delegating this to SenderProvider, not specific senders.

src/lib/notification_service/sender/smtp.py
```diff
@@ -67,12 +60,24 @@ def _get_file_mime_object(file: File) -> MIMEBase:
     def send(self) -> None:
         """Send a message via SMTP."""
         context = ssl.create_default_context()
+        smtp_secret_name = self._delivery_method.get("credentials_secret_name")
 
-        with SMTP_SSL(self._server, self._port, context=context) as server:
+        if smtp_secret_name is None:
+            raise KeyError("Credentials Secret Name is not set.")
+
+        smtp_secret = json.loads(self._secret_client.get_secret(smtp_secret_name))
```
natmasslove, line 68: Maybe, it also make sense to add check if secret's content align with our expectations :) 
Because we ask end-user to create this secret

natmasslove (approved):
I like this implementation. It looks very logical and elegant to me

natmasslove:
Vlad, do you plan any further changes in this PR?

vdektiarev:
> Vlad, do you plan any further changes in this PR?

Sorry, done.

PR #58, opened 2024-02-14
Event_mapper step2 + lambda_alerting pytest coverage

natmasslove:


aryna-yukhnevich (approved):
looks good to me, thank you!

tests/src/test_lambda_alerting.py
```diff
@@ -0,0 +1,325 @@
...
+    
+    event = get_glue_job_event(event_dyn_props_init, "RUNNING", "Job is running")
+
+    result = lambda_handler(event, {})
+
+    assert result["event_is_alertable"] == False, "Event shouldn't raise alert"
+    assert result["event_is_monitorable"] == False, "Event shouldn't be logged"
+    assert (
+        result["resource_type"] == resource_types.GLUE_JOBS
+    ), "Event should be monitorable"
```
vdektiarev, line 121: Not sure I understand the meaning of the comment. event_is_monitorable == False, however we comment that "Event should be monitorable", while it's just a resource type check.

vdektiarev (approved):

tests/src/test_lambda_alerting.py
natmasslove: Thanks!
That's a wicked power of copy-and-paste :)
Fixed

PR #59, opened 2024-02-14
check env variables 

aryna-yukhnevich:
check whether CDK current account and region are aligned with the values provided in the config file


natmasslove (approved):
👍

vdektiarev (approved):

PR #62, opened 2024-02-23
SQS type changed to FIFO

aryna-yukhnevich:
- SQS type changed to FIFO 
- content_based_deduplication enabled*. If we don't enable content-based deduplication and want to deduplicate messages, we would need to provide an explicit deduplication ID in our SendMessage() call. 
https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/FIFO-queues-moving.html

*taking into account that we send messages with unique message bodies

- MessageGroupId attribute added to SQSQueueSender. FIFO queue logic applies only per message group ID. Each message group ID represents a distinct ordered message group within an Amazon SQS queue. For each message group ID, all messages are sent and received in strict order. However, messages with different message group ID values might be sent and received out of order. There is no quota to the number of message groups within a FIFO queue.

https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/FIFO-queues-understanding-logic.html
https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/using-messagegroupid-property.html

- FIFO queue names must end in '.fifo'

image

### DIFF
--- a/infra_tooling_account/infra_tooling_account/infra_tooling_common_stack.py
+++ b/infra_tooling_account/infra_tooling_account/infra_tooling_common_stack.py
@@ -130,7 +130,7 @@ class InfraToolingCommonStack(Stack):
             export_name=AWSNaming.CfnOutput(self, "internal-error-topic-arn"),
         )
 
-    def create_timestream_db(self) -> (timestream.CfnDatabase, kms.Key):
+    def create_timestream_db(self) -> tuple[timestream.CfnDatabase, kms.Key]:
         """Creates Timestream database for events and metrics
 
         Returns:
--- a/infra_tooling_account/infra_tooling_account/infra_tooling_common_stack.py
+++ b/infra_tooling_account/infra_tooling_account/infra_tooling_common_stack.py
@@ -72,6 +72,8 @@ class InfraToolingCommonStack(Stack):
         notification_queue = sqs.Queue(
             self,
             "salmonNotificationQueue",
+            content_based_deduplication=True,
+            fifo=True,
             queue_name=AWSNaming.SQSQueue(self, "notification"),
             visibility_timeout=Duration.seconds(60),
         )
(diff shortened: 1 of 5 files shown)


natmasslove (approved):

infra_tooling_account/infra_tooling_account/infra_tooling_common_stack.py
```diff
@@ -67,11 +67,13 @@ def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
             topic_name=AWSNaming.SNSTopic(self, "internal-error"),
         )
 
-        # Notification SQS Queue
+        # Notification FIFO SQS Queue
         # TODO: confirm visibility timeout
         notification_queue = sqs.Queue(
             self,
             "salmonNotificationQueue",
+            content_based_deduplication=True,
```
vdektiarev, line 75: As I can see, to achieve messages deduplication with _message_group_id_ only, each _message_group_id_ should be unique: https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/using-messagegroupid-property.html#avoding-processing-duplicates-in-multiple-producer-consumer-system

For FIFO queue it makes sense to have message grouping as in our case we don't really care about the ordering, but maybe it's also worth adding a unique **message deduplication ID** to the message to make sure there will be no duplicates regardless of _message_group_id_ ?

Reference: https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/using-messagededuplicationid-property.html

aryna-yukhnevich: yes, right. To achieve messages deduplication we can either provide a unique message deduplication ID for each sent message or enable content-based deduplication 
https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/FIFO-queues-moving.html

At the moment I think that both options can work for us (I assume that we generate the messages with unique content)
 or do you already remember the case when the message content is not unique?

PR #65, opened 2024-03-12
Docs/architecture

natmasslove:


aryna-yukhnevich:
just thinking, should we also add some images with our Grafana dashboards (one with CW logs and at least one with Timestream metrics) so users could have a quick look at them before the deployment?

natmasslove:
> just thinking, should we also add some images with our Grafana dashboards (one with CW logs and at least one with Timestream metrics) so users could have a quick look at them before the deployment?

Removed digest screenshot from an Architecture article. As we discussed, such screenshots should be a part of main readme.md

aryna-yukhnevich (approved):

PR #66, opened 2024-03-13
Fix/glue wf event mapper

natmasslove:


natmasslove:
* fixed Glue WF event mapper init parameters
+ In Glue Workflow notification e-mail now we provide link to execution
image

* fix Glue Workflow notification message - previously email header always showed "COMPLETED" (even for failed executions - default AWS behavior which we override)
* fix events_manager - turns out we can't put more than Events in EventsBus in 1 go

PR #68, opened 2024-03-15
proper handling Glue Data Catalog events + tests coverage

natmasslove:
+ added tests for Glue Data Catalog Event
* Fixed GlueDataCatalog event mapper, so it doesn't fail on create/drop DB / Table / Column operations


aryna-yukhnevich (approved):
looks good to me, thank you!

PR #78, opened 2024-04-05
added mocks for external services in lambda_alerting tests

natmasslove:


natmasslove:
This tiny PR is just to dip my toe's in a muddy water of pytesting :) Specifically, using mocks to ensure no calls to AWS are made.

So, please checks if it's movement in a right direction. Thanks!

natmasslove:
image

Doesn't look to embarrasing :)

tests/src/test_lambda_alerting.py
```diff
@@ -63,6 +63,43 @@ def event_dyn_props_init(aws_props_init):
     return (account_id, region, time_str, epoch_time, version_str)
 
 
+################################################################################################################################
+
+# mocking AWS calls done in lambda_alerting
+
+
+@pytest.fixture(scope="function", autouse=True)
```
npaskrobka, line 71: I would think on the scope, probably you can use a broader scope like "module" or "session" to improve efficiency.
Now scope "function" means mocks will be executed for each test function.

npaskrobka (commented):
Looks very good!
I would advice to test also notification messages returned by lambda (check that result['messages'] expected)

aryna-yukhnevich (approved):

tests/src/test_lambda_alerting.py
```diff
@@ -39,16 +39,16 @@ def aws_props_init():
...
-    ] = f"log-group-salmon-alert-events-{stage_name}"
-    os.environ[
-        "ALERT_EVENTS_CLOUDWATCH_LOG_STREAM_NAME"
-    ] = f"log-stream-salmon-alert-events-{stage_name}"
+    os.environ["ALERT_EVENTS_CLOUDWATCH_LOG_GROUP_NAME"] = (
+        f"log-group-salmon-alert-events-{stage_name}"
+    )
+    os.environ["ALERT_EVENTS_CLOUDWATCH_LOG_STREAM_NAME"] = (
+        f"log-stream-salmon-alert-events-{stage_name}"
+    )
```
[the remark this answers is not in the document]
natmasslove: Good catch. It's no longer needed. Replace with dummy value for stage-name

tests/src/test_lambda_alerting.py
natmasslove: Agree. Changed to module scope.
Had to switch to unittest.mock functionality, as pytest wrapper doesn't allow anything but "function" scope

npaskrobka (approved):

PR #84, opened 2024-04-23
added unit tests for metrics extractors

natmasslove:


natmasslove:
image

Uncovered statements - only those, which doesn't make sense to test (like "pass" or "print("not yet implemented"))

aryna-yukhnevich (approved):

PR #85, opened 2024-04-24
add tests for digest_service

aryna-yukhnevich:
the digest service tests:
image

the coverage:
image

image

missing lines for digest_data_extractor.py   :

- abstract method
image

- not yet implemented methods
image


natmasslove (approved):
Looks very good, thanks!

parametrize approach is very nice. One can argue it's a bit more difficult to read (as opposed to dumb straight-forward approach creating multiple similar tests).
In this case, I believe, parametrize fit in very well, so let's keep it.

One note: there's a type "aggreagted" repeated multiple times in test_digest_data_aggregator. Of course, nothing critical.

aryna-yukhnevich:
Hi @natmasslove ,
Thank you a lot for your review! Typo fixed.

PR #92, opened 2024-06-12
Changed to MainStack + NestedStacks

natmasslove:
Now if grafana settings are removed, relevant stack gets deleted


aryna-yukhnevich (approved):
looks very good :) thank you!

PR #93, opened 2024-06-18
fixes for smtp

aryna-yukhnevich:
- new optional params for smtp delivery method: "use_ssl" (enabled by default), "timeout" (10.0 seconds by default)

- tests:

image

- full code coverage:

image

testing done for STARTTLS option using "smtp.gmail.com" host and 587 port:
image

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -147,7 +147,7 @@ Based on the delivery method type, additional parameters are required:
 * SMTP:
     - `sender_email` - the sender email for notifications and digests.
     - `credentials_secret_name` - The name of the secret stored in AWS Secrets Manager containing the SMTP server credentials. Required key-value pairs: `SMTP_SERVER`, `SMTP_PORT`, `SMTP_LOGIN`, `SMTP_PASSWORD`.
-    - (optional) `use_ssl` - indicate whether to use SSL for the SMTP server connection. If set to True, the connection will use SSL. Otherwhise STARTTLS will be used. Default value: `True`.
+    - (optional) `use_ssl` - indicate whether to use SSL for the SMTP server connection. If set to True, the connection will use SSL. Otherwhise, STARTTLS will be used. Default value: `True`.
     - (optional) `timeout` -  the connection timeout in seconds. Default value: `10.0`.
 
 You can specify multiple delivery methods (even for the same delivery type, no restrictions).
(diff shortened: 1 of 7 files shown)


docs/configuration.md
```diff
@@ -144,6 +144,11 @@ Based on the delivery method type, additional parameters are required:
     - `sender_email` - the sender email for notifications and digests. Must be verified in AWS SES.
 * AWS_SNS:
     - No additional parameters needed. Target SNS topic Arn is configured in recipients section.
+* SMTP:
+    - `sender_email` - the sender email for notifications and digests.
+    - `credentials_secret_name` - the name of the secret stored in AWS Secrets Manager containing the SMTP server credentials. Required key-value pairs: `SMTP_SERVER`, `SMTP_PORT`, `SMTP_LOGIN`, `SMTP_PASSWORD`.
```
natmasslove, line 149: just as an idea - maybe we can also give sample json context of the secret
{ "SMTP_SERVER" : "put_your_host_here",
  "SMTP_PORT" ...
}

infra_tooling_account/infra_tooling_account/infra_tooling_common_stack.py
```diff
@@ -249,6 +251,16 @@ def create_notification_lambda(
...
         )
 
+        notification_lambda_role.add_to_policy(
+            iam.PolicyStatement(
+                actions=["secretsmanager:GetSecretValue"],
+                effect=iam.Effect.ALLOW,
+                resources=[
+                    "*"
+                ],  # to be able to retrieve Secrets required for SMTP sender
+            )
```
natmasslove, line 261: maybe, an idea for future - instead of allowing getsecretvalue for "*"
we can:
1) ask user put a specific tag on SMTP secrets
2) role will be able to read "*" resources, but only those which are also tagged "config" : "salmon", for example.
Let's discuss in the next sync-up

natmasslove (approved):
Look good to me! Thanks!

docs/configuration.md
aryna-yukhnevich: good idea, I've added this sample, thank you!

infra_tooling_account/infra_tooling_account/infra_tooling_common_stack.py
aryna-yukhnevich: sure, I support this idea, thank you!

PR #94, opened 2024-06-19
1. Switched to built-in CDK way of passing resource references between stacks (instead of CfnOutput, Fn:ImportValue)

natmasslove:


natmasslove:
2. Moved relevant outputs to Main Stack (settings bucket name and Grafana Instance URL / Secret - when aplicable)

3. Updated Grafana-related doc

docs/changelog.md
```diff
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file.
 * Unit test code coverage improved to ~90%
 * Notifications
   * Added AWS_SNS delivery method type
+* Improved CDK code (for tooling account part) readability and manageability by switching to Nested Stacks approach.
```
aryna-yukhnevich, line 10: should we mention in our docs about the upgrade process (from ver1.0 to 1.1)? or is it just about running the cdk deploy command?

aryna-yukhnevich (approved):
looks really good to me :)

docs/changelog.md
natmasslove: Usual cycle of cdk destroy / cdk deploy should be sufficient here, so I guess we can leave the update statement "as is".

PR #96, opened 2024-06-21
Feature/add account id to s3 bucket name

aryna-yukhnevich:
image

### DIFF
--- a/infra_tooling_account/infra_tooling_account/infra_tooling_common_stack.py
+++ b/infra_tooling_account/infra_tooling_account/infra_tooling_common_stack.py
@@ -112,7 +112,9 @@ class InfraToolingCommonStack(NestedStack):
         settings_bucket = s3.Bucket(
             self,
             "salmonSettingsBucket",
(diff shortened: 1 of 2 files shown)


infra_tooling_account/infra_tooling_account/infra_tooling_common_stack.py
```diff
@@ -112,7 +112,9 @@ def create_settings_bucket(self) -> s3.Bucket:
         settings_bucket = s3.Bucket(
             self,
             "salmonSettingsBucket",
-            bucket_name=AWSNaming.S3Bucket(self, "settings"),
+            bucket_name=AWSNaming.S3Bucket(
+                self, "settings", NestedStack.of(self).account
```
natmasslove, line 116: I'm wondering if it might be easier just pass
"settings-" + account_id as meaning param value and not to change awsnaming param list?
What do you think?

aryna-yukhnevich: sure, good idea, let me adjust it
thank you!

PR #103, opened 2024-07-18
add dq metrics extractor

aryna-yukhnevich:
- DQ metrics extractor added
- Unit tests added
 image
- Total coverage = 91%
image
- data from Timestream table:
image

- dashboard added
image

- digest
image


docs/changelog.md
```diff
@@ -3,7 +3,7 @@
 All notable changes to this project will be documented in this file.
 
 ## 1.2.0
-* Alerting functionality implemented for Glue Data Quality events
+* Added Glue Data Quality service (alerts, metrics, digest, optional Grafana dashboard).
```
natmasslove, line 6: let's maybe rephrase a bit. to something like "Added integration with AWS Glue Data Quality (..."

docs/configuration.md
```diff
@@ -203,7 +203,8 @@ Within each section, list the resources of the corresponding resource type along
 
     - `name` - specify the resource name to be monitored.
 
-        >  If you would like to monitor the resources with a common pattern in their names (e.g., glue-pipeline1-ingest, glue-pipeline1-cleanse, glue-pipeline1-staging), use wildcards: glue-pipeline1`-*`.
+        >  If you would like to monitor the resources with a common pattern in their names (e.g., glue-pipeline1-ingest, glue-pipeline1-cleanse, glue-pipeline1-staging), use wildcards: glue-pipeline1`-*`. \
+        > Note: Wildcards are not supported for Glue Data Quality Rulesets executed within Glue Jobs. The full name of the Ruleset should be provided.
```
natmasslove, line 207: not critical, but I'm wondering if we change to "positively formulated" statements.
Like 
Note: In Glue DQ wildcards are supported only for Rulesets executed in .... For other Rulesets please specify an exact name (without wildcards).
Again, not critical, just polishing.

src/lib/metrics_extractor/metrics_extractor_utils.py
```diff
@@ -109,3 +110,26 @@ def get_last_update_time(
             return datettime_utc
 
     return None
+
+
+def get_min_update_time(
```
natmasslove, line 115: Function name might become a bit misleading in the future, I guess.
As it searches for the min_update_time AMONG last_update_time.

So, maybe get_earliest_last_update_time or get_earliest_last_update_time_for_resource_set might be a bit more descriptive. What do you think?

docs/configuration.md
aryna-yukhnevich: sure, will update accordingly

src/lib/metrics_extractor/metrics_extractor_utils.py
aryna-yukhnevich: yes, totally agree with your point

src/lib/metrics_extractor/base_metrics_extractor.py
```diff
@@ -72,12 +71,15 @@ def get_last_update_time(self, timestream_query_client) -> datetime:
         return last_date
 
     @abstractmethod
-    def prepare_metrics_data(self, since_time: datetime) -> (list, dict):
+    def prepare_metrics_data(
+        self, since_time: datetime, result_ids: list = []
```
natmasslove, line 75: I'd consider another way of implementing an additional parameter, which is specific to Glue DQ.

1. don't add a parameter to this method (so you won't need to change all other classes implementation)
2. in glue DQ metrics extractor add set_results_id  (self.results_ids = results_ids)
3. in glue DQ's prepare_metrics_data implementation refer to self.results_id
4. in lambda_extract_metric add something like 
if Glue DQ:
   extractor.set_result_ids
   
   Hopefully, this makes sense. Anyway, it'd be nice avoid changing all other classes implementation just for one-class specific parameter

aryna-yukhnevich: sounds good to me, thanks a lot!

natmasslove (approved):

PR #105, opened 2024-07-26
Changed Deployment-Tests workflow to use composite actions

natmasslove:
this enables future decoupling of cdk prepare / destroy / deploy actions for integrational tests


aryna-yukhnevich (approved):
looks good to me, thanks!

PR #108, opened 2024-08-01
Feature/integration tests v1

natmasslove:
PR goals:
- create a pipeline which provisions infra for integration tests (tests are not yet triggered automatically)
- make sure target SQS queue (where we expect to see all events during tests) is populated:
either when monitored resource failed or internal error occured (e.g. salmon lambda failed)
  this was tested manually
- prepare initial testing stand (with just 2 glue jobs - 1 always fails, another - always succeeds)
- prepare salmon settings to cover these jobs + recipient (target SNS topic which, eventually, sends everything into target SQS queue)


aryna-yukhnevich (approved):
looks good to me, thank you!

PR #112, opened 2024-08-06
Integration Tests - Glue Jobs (part 1)

natmasslove:
1. split IntegrationTesting github workflows into DeployInfra + ExecuteTests + DestroyInfra

so ExecuteTests can be ran multiple times without infra re-provisioning

2. added GlueRunner class + BaseRunner (which will be inherited also for other services)
runner classes are responsible for launching testing stand resources in async mode and waiting for their completion

3. added classes to 
retrieve Sqs messages, 
filter them by timestamp (so only messages relevant to recent testing stand execution are in scope)
filter them by criteria (e.g. subject contains "FAILED" and "glue_jobs")

4. tested pipeline execution (using temp credentials as, so far, github_service_user lacks some privileges)

todo for part 2:
2. implement "full" pipeline (which orchestrates DeployInfra + ExecuteTests + DestroyInfra )
3. add tests for timestream records checks (+ classes to retrieve those)
5. IAM Role - add permissions and finally switch back to github_service_user

image


.github/workflows/integration_tests_infra_delete.yml
```diff
@@ -0,0 +1,52 @@
...
+        - name: prepare CDK
+          uses: ./.github/actions/cdk-prepare
+          with:
+            cdk-folder-name: ${{ env.CDK_INTEGRATION_TESTS_FOLDER }}
+            settings-folder-name: ${{ env.SETTING_FOLDER_NAME }}
+            stage-name: ${{ env.STAGE_NAME }}
+            region: ${{ env.AWS_DEFAULT_REGION }}
+
+        # 1. Destroy Monitored env
+        - name: deploy monitored infra via CDK
```
aryna-yukhnevich, line 29: just a small comment on the step names - should we name them as 'destroy...'?

PR #113, opened 2024-08-09
Intergration tests for Glue Jobs + foundation: part2

natmasslove:
+ added tests checking timestream records are populated correctly
+ added COMMON test to check if digest has arrived
+ added glue-job specific test to check digest content - verify it contains mentions of all glue jobs in scope

* fixed deployment pipeline (wrong lib reference)


aryna-yukhnevich (approved):
looks good to me :)

PR #115, opened 2024-08-13
Feature/add emr serverless metrics extractor

aryna-yukhnevich:
- emr data in the Timestream table
image

- unit tests added
image
image

- the dashboard added
image

- small fix applied to get_earliest_last_update_time_for_resource_set to handle the case when last_update_times is None

- some updates in the digest (added RESOURCE_TYPE_DECORATED_NAMES constants to display Service name and Resource type name appropriately; error message extended to 100 chars)

image

### DIFF
--- a/src/lib/digest_service/digest_data_aggregator.py
+++ b/src/lib/digest_service/digest_data_aggregator.py
@@ -94,14 +94,20 @@ class DigestDataAggregator:
             )
 
             # construct error comment based on the job run URL and error message
+            error_message = resource_run.get("error_message", "Unknown errors")
             if job_run_url:
-                error_comment = f" - <a href={job_run_url}>ERROR: {resource_run['error_message'][:50]}</a>"
-            elif resource_run["error_message"]:
-                # if no job run, use the error message directly
-                error_comment = resource_run["error_message"]
+                # truncate error message if > 100 chars
+                truncated_message = (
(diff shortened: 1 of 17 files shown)


src/lib/digest_service/digest_data_aggregator.py
```diff
@@ -94,14 +94,20 @@ def _get_error_comments(
...
             # construct error comment based on the job run URL and error message
+            error_message = resource_run.get("error_message", "Unknown errors")
             if job_run_url:
-                error_comment = f" - <a href={job_run_url}>ERROR: {resource_run['error_message'][:50]}</a>"
-            elif resource_run["error_message"]:
-                # if no job run, use the error message directly
-                error_comment = resource_run["error_message"]
+                # truncate error message if > 100 chars
+                truncated_message = (
+                    (error_message[:100] + "...")
```
natmasslove, line 101: maybe, it make sense to define max error length text as a constant (if though we still can keep the constant in this module)

natmasslove (approved):
Looks cool, thanks!

src/lib/digest_service/digest_data_aggregator.py
aryna-yukhnevich: yes, sure. also had this idea in mind

PR #116, opened 2024-08-13
Added GitHub actions service user

natmasslove:
1. Github actions service user credentials are used in GitHub workflows (CDK deployment tests, Integration tests)
2. If you choose to use integration tests together with Salmon - you need to deploy CDK app for github actions resources (currently, it's IAM user only + it's privileges).
3. Privileges granted IAM user are minimally sufficient for:
- deploying / destroying integration and deployment tests infra
- executing integration tests (interaction with testing stand resources to launch them + working with tooling env to extract metrics / analyze alerts )
4. Restructured CDK part of the project.
Created a separate folder "cdk" where all CDK Salmon apps should reside (some app are yet to be moved there)
5. Integration tests are designed in a way, so they:
- can be executed "all steps in one" meaning create infra->execute tests->delete infra (Full IntTest workflow)
- can be executed separately (e.g. create infra once, but execute tests multiple time - separate workflows)

- can be executed via GitHub actions, but also from local computer, e.g.:
`python integration_tests/testing_stand_execution.py --stage-name=devit --region=eu-central-1`
`pytest integration_tests/tests --stage-name=devit --region=eu-central-1 --start-epochtimemsec=1723548392000`
(start-epochtimemsec value should be taken from testing_stand_execution.py)

### DIFF
--- a/.github/workflows/integration_tests_execute.yml
+++ b/.github/workflows/integration_tests_execute.yml
@@ -9,7 +9,9 @@ env:
     STAGE_NAME: devit
     SETTING_FOLDER_NAME: "integration_tests/settings"
 run-name: integration tests - execute tests
-on: [workflow_dispatch]
+on: 
+  workflow_dispatch:
+  repository_dispatch:
 jobs:
   integration-testing:
     runs-on: ubuntu-latest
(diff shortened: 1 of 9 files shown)


aryna-yukhnevich (approved):
looks good to me, thanks!

PR #118, opened 2024-08-19
Moved CDK tooling and monitored env code to cdk/ folder

natmasslove:


aryna-yukhnevich (approved):
looks good to me, thank you!

PR #119, opened 2024-08-19
fix the issue in get_monitoring_groups

aryna-yukhnevich:
- extend get_monitoring_groups with resource type so to avoid an incorrect group assignment in case "*" specified in the config file as a resource name to be monitored
- add a unit test test_get_monitoring_groups_wildcards to cover the case mentioned above


natmasslove (approved):
Looks very good, thanks!

PR #130, opened 2024-09-10
restructured doc article for testing + documented --resource-types param

natmasslove:


aryna-yukhnevich (approved):
looks very good to me, thanks!

PR #135, opened 2024-09-18
fix the issue with getting lambda logs since the last update time

aryna-yukhnevich:
- query start time for getting lambda logs was increased by 1millisecond so to make since_time passed non-inclusive

### DIFF
--- a/integration_tests/inttest_lib/runners/lambda_function_runner.py
+++ b/integration_tests/inttest_lib/runners/lambda_function_runner.py
@@ -1,12 +1,9 @@
 import boto3
 import time
 import json
-from datetime import datetime, timedelta
 
 from inttest_lib.runners.base_resource_runner import BaseResourceRunner
-
 from lib.aws.cloudwatch_manager import CloudWatchManager, CloudWatchManagerException
-from lib.core.datetime_utils import datetime_to_epoch_milliseconds
 
 
 class LambdaFunctionRunnerException(Exception):
@@ -23,6 +20,7 @@ class LambdaFunctionRunner(BaseResourceRunner):
         self.function_runs = {}
 
     def initiate(self):
+        self.start_time = int(time.time() * 1000)
         for lambda_function_name in self.resource_names:
             request_id = self._invoke_lambda_async(lambda_function_name)
             self.function_runs[lambda_function_name] = request_id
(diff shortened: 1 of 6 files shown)


integration_tests/inttest_lib/runners/lambda_function_runner.py
```diff
@@ -74,8 +75,10 @@ def _is_lambda_completed(self, lambda_function_name, request_id, start_time):
         query_string = f"fields @timestamp, @message | filter @message like /END RequestId: {request_id}/"
 
         try:
+            query_start_time = int(datetime_to_epoch_milliseconds(start_time))
```
natmasslove, line 78: can you please double-check it doesn't break LambdaFunctionRunner in integration tests (it also uses query_logs)

natmasslove (approved):
Look good to me, thank!
We just need to address a concern regardging LambdaFunctionRunner

integration_tests/inttest_lib/runners/lambda_function_runner.py
natmasslove, line 78: since we are relying on int, which is epoch msec, I'd rather used int(time.time()*1000) directly as it provides UTC, tz-aware value.
maybe, we can change start_time assignment in await_completion method.
Can discuss during the next sync.

p.s.: removed my previous comment as it didn't make sense :)
but this, hopefully, does

aryna-yukhnevich: thank you for your comments!
sure :) let me try to check on that as well

integration_tests/tests/test_lambda_functions.py
```diff
@@ -24,16 +27,21 @@ def lambda_execution_timestream_metrics_summary(region, start_epochtimemsec, sta
...
                 FROM "{DB_NAME}"."{TABLE_NAME}"
                 WHERE measure_name = 'execution' AND time > from_milliseconds({start_epochtimemsec})
-    """   
+    """
     result = query_runner.execute_query(query=query)
 
-    return result[0] # returning the first record (it's only 1 record in resultset by query design)
+    return result[
+        0
+    ]  # returning the first record (it's only 1 record in resultset by query design)
```
natmasslove, line 35: if we put comment on a line before (not the same one), formatting would be less weird :)

aryna-yukhnevich: noted, thank you :)

PR #138, opened 2024-09-23
Integration tests/unique stage name

natmasslove:
IntTest envs now named named as "it{first 3 letters of authors name}"
for "main" branch, the name is "itmain".


aryna-yukhnevich (approved):
I like this approach, thank you!

PR #139, opened 2024-09-25
Fix/lambda logs issue

natmasslove:
LambdaRunner - added explicit log_group creation when needed


aryna-yukhnevich (approved):
looks good to me, thanks!

PR #146, opened 2024-10-15
add one more Lambda example to integration testing stand

aryna-yukhnevich:
- lambda created 
image

- tests coverage
image

- integration testing finished successfully

image

- ResourceTypeResolver updated to handle only relevant events


src/lambda_alerting.py
```diff
@@ -55,6 +55,12 @@ def lambda_handler(event, context):
     settings = Settings.from_s3_path(settings_s3_path)
 
     resource_type = ResourceTypeResolver.resolve(event)
+    if not resource_type:
+        logger.info(
+            f"Event for the source {event['source']} and detail-type {event['detail-type']} is not alertable."
```
natmasslove, line 60: maybe we can rephrase from "is not alertable" to something like "are not handled by SALMON"?
To distinguish those from events which we can process, but by business logic they are not-alertable?

natmasslove (approved):
with minor comment

src/lambda_alerting.py
aryna-yukhnevich: sure, thank you!

PR #147, opened 2024-10-15
fix get_all_names for all services (paginator / next token)

natmasslove:
modified get_all_names for all services, so if result doesn't fit in one default page, it would use paginator or next token to retrieve full list.


aryna-yukhnevich (approved):
looks good to me, thank you!

PR #149, opened 2024-10-20
Integration tests/purge cw logs lambdas

natmasslove:
When integration tests infra is deleted, CloudWatch logs for testing stand lambdas are purged (so they won't result in fake error messages from previous executions)


aryna-yukhnevich (approved):
looks very good to me, thanks !

PR #154, opened 2024-11-05
OpenID authentication for GitHub workflows

natmasslove:
1. added IAM service role which is now used by GitHub actions (instead of IAM service user).
2. service user credentials are no longer stored in git repo's secrets
3. Role permissions are identical to those were used for service user
4. github_actions_resources CDK stack still maintains IAM service user, so we can have credentials to run integration tests locally with the same level of permissions as github actions have.


aryna-yukhnevich (approved):
Great job, thank you!

PR #155, opened 2024-11-08
new doc on how to integrate new AWS service added

aryna-yukhnevich:


README.md
```diff
@@ -119,4 +120,5 @@ If you plan to contribute to the project, please refer to [tests overview articl
 ## Other useful links
 
 * [FAQ & Troubleshooting](docs/faq.md)
-* [Version history](docs/changelog.md)
\ No newline at end of file
+* [Version history](docs/changelog.md)
+* [Steps to Intregrate New Service](docs/service_integration.md)
```
natmasslove, line 124: One important thing here is to point out that the info is relevant for contributors only, not end-users.

I guess, it better fits into "Contributing" section of the documentation (instead of general-purpose "Other useful links" section). In "contributing" we can say - if you want to integrate ..., then - follow this guide.

docs/service_integration.md
```diff
@@ -0,0 +1,46 @@
+# SALMON New Service Integration
```
natmasslove, line 1: I would also consider slightest changes in wording (in .md file name and inside the doc).
like "adding_aws_service_integration.md"
And "AWS" from my perspective might be important.

README.md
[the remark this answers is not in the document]
aryna-yukhnevich: oh yeah ,sorry for that

natmasslove (approved):
Overall, looks very good, thanks!
Let's just consider improving the logical structure and wording in several places.

PR #156, opened 2024-11-12
update IAM policy to limit access to SNS topics tagged with salmon

aryna-yukhnevich:


natmasslove (approved):
Looks very good, thanks!

PR #158, opened 2024-11-17
Adding capabilities to use limited permission CDK exec role (instead of default one)

natmasslove:
1. created IAM policy to replace default AdministratorAccess policy for CDK execution role
2. updated / added documentation on how to create and use this policy
3. modified CDK Deployment tests workflow to make sure it always uses "limited" policy

It doesn't make sense to apply policy to integration test, as those are not customer-related artifacts.


aryna-yukhnevich (approved):
looks good to me, thank you!

PR #169, opened 2024-12-19
Feature/metrics storage step1

natmasslove:


src/lib/metrics_extractor/base_metrics_extractor.py
```diff
@@ -54,26 +51,8 @@ def get_aws_service_client(self, aws_client_name: str = None):
...
-        ):
-            return None
-
-        query = f'SELECT max(time) FROM "{self.timestream_db_name}"."{self.timestream_metrics_table_name}" WHERE {self.RESOURCE_NAME_COLUMN_NAME} = \'{self.resource_name}\''
-        last_date = queryRunner.execute_scalar_query_date_field(query=query)
-        return last_date
-
     @abstractmethod
-    def prepare_metrics_data(self, since_time: datetime) -> (list, dict):
+    def prepare_metrics_data(self, since_time: datetime) -> (list, dict):  # type: ignore
```
aryna-yukhnevich, line 55: just for my understanding, could you please elaborate on the benefits of using # type: ignore here instead of annotating as tuple[list, dict]?

natmasslove: no benefits :) fixed. good catch, thanks.

aryna-yukhnevich (approved):
Looks very good to me, thank you!

PR #171, opened 2024-12-31
metrics storage refactor. step3

natmasslove:
1. Refactor digest (lambda, extractor, extractor_provider) to use metrics_storage, but not TimestreamQueryRunner directly
2. Renamed lambda config env params TIMESTREAM_METRICS_DB_NAME -> METRICS_DB_NAME
3. Upgraded aws_cdk_lib 2.103 -> 2.174 (required to use PYTHON_3_13 in lambdas).
4. Upgrading aws_cdk_lib broke CDK monitored stack (Error: Cannot use a service or account principal with grantAssumeRole, use assumeRolePolicy instead.)
5. Now digest_extract_provider uses TypedDict for **kwargs


aryna-yukhnevich (approved):
looks very good to me, thank you!

